PR #4, opened 2016-02-10
adding matplotlib inline for notebook

ajkl:


JosPolfliet:
pandas-profiling should work without the matplotlib inline command. If that is not the case, please let me know. If this is just something you want to add in a personal notebook, go ahead but I won't merge it to this example notebook.

ajkl:
It did not work for me without the inline.

JosPolfliet:
What version of matplotlib, pandas, Python and Jupyter are you using? I'll try to recreate

ajkl:
matplotlib - 1.5.0
pandas - 0.17.1
python - 2.7.11
notebook server - 4.0.6

JosPolfliet:
Couldn't replicate it so it does not look like a dependency issue. Can you add the output of `pip freeze` and the full traceback of the error you are seeing? If not an error, can you add a screenshot of what is wrong? What OS/browser are you using?

PR #17, opened 2016-05-31
Function to make scatter plot

brendabrandy:
From the issues forum, @JosPolfliet suggested that it would be nice to have a scatter plot to show correlations between data. I made a scatter function that prepares a dictionary for a scatter plot if both categories are considered as numerical. I will work more on it later if you think it's good :) And if I move forward with it, how should I work it out? Do i need javascript to make the scatter plot? Or HTML and CSS would be fine?


brendabrandy:
Oops, I think there's a mistype on the return line. Please change if pull request is accepted. Thanks!

JosPolfliet:
Hi @brendabrandy ,

The histogram plots are created in the mini_histogram(series) function inside the describe function. They use matplotlib as a backend for the plotting and then import that image in the HTML of the report. 

If you want to add a scatter plot I would propose keeping the same format. The image should be converted to a base64 encoded string which is a part of the description dictionary (e.g. key 'scatter') that is returned by the describe() function. While converting all plots to JS using a library would have some benefits, I am afraid it would also significantly impact performance.

For the scatter plots, we would have to define a way of specifying a target variable as well. I propose adding a target= keyword parameter in the describe function. If there is interest later, we can even to some preliminary modelling to determine contributing variables and add correlation statistics with the target variable for each explanatory variable.

Thanks for contributing! Once the whole flow works I will review and merge.

Jos

PR #19, opened 2016-07-14
added optional corr_threshold and display warnings toggle

bartlesy:
The corr_threshold parameter can be used to set the cutoff for variables to be analyzed by the profiler, e.g. a corr_threshold of 1 will analyze all variables regardless of any high correlations pairs, and a corr_threshold .5 will only analyze one variable in a pair of variables with correlation > .5.

I added a toggle to display warnings to html generated output:
image

vs

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ target/
 
 # remove pycharm boilerplate files
 .idea
+SB_PUSH_INSTRUCTIONS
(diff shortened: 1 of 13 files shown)


conradoqg:
Guys,

This fork took a different path and most of the commits in the master is now related to its internal release see [here](https://github.com/JosPolfliet/pandas-profiling/pull/19/files/64e758f09126ace060e747bf85d8dd65dcc8760f..10ea9e1a1f88e147be2f556e82d72b9f2a55396d)

I think we should close this PR.

Best

PR #20, opened 2016-08-04
Add correlation overrides to make sure to get data for fields even if…

liorregev:
Add correlation overrides to make sure to get data for fields even if the are highly correlated


JosPolfliet:
Hi  @liorregev, I merged your changes manually since your branch was behind dozens of commits. I also removed the copy_reg dependency. The multi core support is a great addition by the way, pandas-profiling is a lot faster now! Thanks for contributing!

Jos

PR #32, opened 2016-12-15
added call to pool.close() in describe_1d

kevanshea:
Thanks for this module. It's saved me a lot of time in working on a script for profiling databases.

In running this inside another python script on multiple tables in a database I would encounter "too many files open" error after it processed enough columns. The problem seems to be the threads were staying open and in turn keeping a file descriptor open. Once enough threads had been started it would hit the OS's limit on open files.

This additional line should fix the problem.


JosPolfliet:
Good catch, thanks! Somehow I assumed the pool would close at the end on the method. Not sure why.

PR #43, opened 2017-04-21
Add keyword arg to not check correlation.

lawlesst:
Thanks for this package. We are evaluating it for project and finding it suits our needs. 

In our use case we want to profile rather large database tables. We found that the correlation checks can use a lot of memory causing the process to fail. Since this check isn't critical to this use case, we would simply like to skip these checks when profiling. 

This pull request adds a keyword argument to `describe` called `check_correlation` that defaults to `True`. If `False` is passed, then the correlation checks aren't run.

### DIFF
--- a/pandas_profiling/base.py
+++ b/pandas_profiling/base.py
@@ -247,11 +247,12 @@ def multiprocess_func(x, **kwargs):
     return x[0], describe_1d(x[1], **kwargs)
 
 
-def describe(df, bins=10, correlation_overrides=None, pool_size=multiprocessing.cpu_count(), **kwargs):
+def describe(df, bins=10, check_correlation=True, correlation_overrides=None, pool_size=multiprocessing.cpu_count(), **kwargs):
     """
     Generates a object containing summary statistics for a given DataFrame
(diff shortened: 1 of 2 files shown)


Yashg19:
Do you know when will the changes for "check correlation kwarg to Profile report fails #62 " will be moved to production (i.e. to PyPI release)?

gaswani:
When I execute : pandas_profiling.ProfileReport(df), I am not getting the Pearson and Spearman correlation matrices in the output report. In the overview the correlation matrices are missing. I would like to have these two in the overview. How do I resolve this problem? Someone please help.

Thanks in advance.

sbrugman:
Hi @gaswani, this issue is resolved in the latest version of `pandas-profiling`, however this is not yet available at PyPi. As a temporary solution, you can obtain the latest version via `pip install git+https://github.com/pandas-profiling/pandas-profiling.git`.

gaswani:
Thanks @sbrugman Let me try that out for now

gaswani:
@sbrugman What/which is the latest version?

Issue #11, opened 2016-04-25
Difficulty getting access to latest `get_rejected_variables()` functionality


proinsias:
I installed `pandas-profiling` using:

`pip install pandas-profiling`

This gave me `pandas-profiling 1.0.0a2`, but the corresponding `__init__.py` file did not contain the `get_rejected_variables()` functionality. 

I then cloned the `git` repo, and tried to install using:

`python setup.py install`

This gave me the zipped egg file in `~/anaconda3/envs/python2/lib/python2.7/site-packages`, and I got an error that python could not find `pandas_profiling.mplstyle` because the `pandas_profiling` directory did not exist.

Finally, I extracted the egg file and move `pandas_profiling` into `~/anaconda3/envs/python2/lib/python2.7/site-packages`, and got the new functionality.

Might be something I missed. Just letting you know.

JosPolfliet:
That functionality is not committed to PyPi yet. Will do so over the weekend. 

The zipped egg should contain the mplstyle file. I will look into it and try to recreate... stay tuned...

beaunorgeot:
Greetings! Great project concept

Is this complete yet? 
calls to get_rejected_variables() return an error for me:
AttributeError: 'DataFrame' object has no attribute 'correlation'

All other pandas_profiling functionality seems to work for me. 

replicate:
import pandas as pd
import pandas_profiling
import numpy as np

ufo = pd.read_table('http://bit.ly/uforeports', sep=',', parse_dates=['Time'])
profile = pandas_profiling.ProfileReport(ufo)
rejected_variables = profile.get_rejected_variables(threshold=0.9)

romainx:
The bug has been fixed by @conradoqg and merged on the master branch.
Regards.

PR #79, opened 2017-12-31
Instead of throwing an error check if the correlation was computed

conradoqg:
This fixes issue #11. In his example, the dataset doesn't compute the correlation, so if the attribute doesn't exist, we should return an empty array.


romainx:
Hello,

Thanks a lot, it makes sense.

Issue #77, opened 2017-12-15
ProfileReport fails if dataframe contains boolean columns


Aylr:
Despite #25 and #64  , this appears to still be an issue.

## Steps to reproduce

```
foo = pd.DataFrame({
    'a': [1, 2, 3],
    'bool': [True, False, True],
})

# fails
pandas_profiling.ProfileReport(foo)

# succeeds (convert to integer)
foo['bool'] = np.where(foo['bool'], 1, 0)
pandas_profiling.ProfileReport(foo)
```

Does anyone have any suggestions before I submit a PR?

conradoqg:
I wasn't able to reproduce this problem with the latest commit.

Could you give more details about your environment?

My:
```
CPython 3.6.0
IPython 6.2.1

pandas 0.19.2
numpy 1.13.3
matplotlib 2.1.0
six 1.11.0

compiler   : MSC v.1900 64 bit (AMD64)
system     : Windows
release    : 10
machine    : AMD64
processor  : Intel64 Family 6 Model 94 Stepping 3, GenuineIntel
CPU cores  : 8
interpreter: 64bit
Git hash   : bc52a3cc10863c54c592210e86989d3ef733f138
```

Best

romainx:
I confirm, I'm not able to reproduce the problem in my environment.
Could you also share the error observed?

Thanks

Aylr:
I was not able to reproduce this! When I was certain I could it turned out ipython must have been caching a non-master branch of the library.

From (another) clean environment I was able to get both code snippets above to work.

Sorry about the user error!!

Aylr:
Since it's been almost a year since the last release when can we expect a new one on pypi? :smile:

conradoqg:
Hey,

I think a new release is a good idea but before that, this month I hope to be able to do a series of new features and bug fixes.

If the authors can wait a week, I think the most important ones will be ready.

Best

romainx:
Hello,

I'm not the author, but I agree. I think new features are mandatory before delivering a new version and I will help improving it. On my side I would be happy to improve testing and CI. 

We have to check with @JosPolfliet his target date. In my point of view we should commit to issue a new version before the end of the first quarter 😄.

conradoqg:
Hey,

I'm confident that after the PR #82 we don't have any more problems with bool variables.

@romainx Can you open a specific issue to release a new version (maybe we can create a list of desired improvements/issues to be fixed).

I think we can close this issue after those changes.

[...]

Issue #76, opened 2017-12-04
ProfileReport extremly slow on a 40K lines file


MordicusEtCubitus:
I'm using an CSV file from opendata.
Data is not perfectly clean, but is used without issue with pandas.
When I try to run pandas_profiling.ProfileReport() on it, jupyter goes into a very very long run using lot of cpu and never giving back hand.

File is not very big, only 36700 records. I've attached it to this defect.
If you can help...
Thanks a lot, and also for your nice idea !

[villes_france.csv.zip](https://github.com/JosPolfliet/pandas-profiling/files/1528941/villes_france.csv.zip).

conradoqg:
Hey,

Two things are happening in this case:
- The "check correlation" feature of this profiling tool is very heavy due the check of correlation over all combinations of categorical features. If you deactivate that feature `pfr = pandas_profiling.ProfileReport(df, check_correlation=False)` it will not crash due to memory usage, but it will throw an error, that I explain below.
- This tool doesn't support multiple dtypes for the same column, and to solve this problem the dataset must be cleaned beforehand or you could force the dtype `df = pd.read_csv("../examples/villes_france.csv", dtype={'dep': str, 'code insee': str}, encoding='UTF-8')`.

Working version of this case:

```
import pandas_profiling
import pandas as pd

df = pd.read_csv("../examples/villes_france.csv", dtype={'dep': str, 'code insee': str}, encoding='UTF-8')
pfr = pandas_profiling.ProfileReport(df, check_correlation=False)
pfr.to_file("/tmp/example.html")
```

I will look if there is anything I can do solve those problems or at least inform the user what's happening.

conradoqg:
Hey,

With the above PR #86 merged, you can:
- Use the parameter `pandas_profiling.ProfileReport(df, check_recoded = False)` to alleviate memory pressure and at the same time get the correlation report. I think this is the most we can do right now. The recoded checking is indeed memory heavy.
- You don't need to force the dtype anymore, the ProfileReport will automatically ignore unsupported types.

I think we can close this issue after those changes.

Best

romainx:
Hello,

Thanks for the update. I close the issue. 
We will reopen it if needed.

RiteshBabel:
facing Value error:
Config parameter "check_correlation" does not exist.
after apling pandas_profiling.ProfileReport(df ,check_correlation=False)

TylerWakefield:
also getting the error above

BryTyr:
[...]

Issue #70, opened 2017-11-13
Properly handle columns with mixed numeric and categorical values


navado:
You might want to handle them as categorical or define parameter to solve the following problem:
```
  File "/home/user/.local/lib/python3.5/site-packages/pandas_profiling/__init__.py", line 17, in __init__
    description_set = describe(df, **kwargs)
  File "/home/user/.local/lib/python3.5/site-packages/pandas_profiling/base.py", line 308, in describe
    confusion_matrix=pd.crosstab(data1,data2)
  File "/home/user/.local/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/core/reshape/pivot.py", line 493, in crosstab
    aggfunc=len, margins=margins, dropna=dropna)
  File "/home/user/.local/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/core/reshape/pivot.py", line 160, in pivot_table
    table = table.sort_index(axis=1)
  File "/home/user/.local/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/core/frame.py", line 3243, in sort_index
    labels = labels._sort_levels_monotonic()
  File "/home/user/.local/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/core/indexes/multi.py", line 1241, in _sort_levels_monotonic
    indexer = lev.argsort()
  File "/home/user/.local/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/core/indexes/base.py", line 2091, in argsort
    return result.argsort(*args, **kwargs)
TypeError: unorderable types: str() < int()
```

romainx:
Please could you provide an example permitting to reproduce the case?

Note: Since #43, it is possible to call the report with the option `check_correlation=False` to skip this phase. In your case, it will avoid the error since it appears during the correlation computing.

navado:
Example data: [ex.txt](https://github.com/JosPolfliet/pandas-profiling/files/1469977/ex.txt)
More observations - if I reduce the size of the example to ~1000 points, including the problematic records, the problem does not happen. 
Another strange thing, enabling `check_correlation=False` throws an error
```
>>> pfr = pandas_profiling.ProfileReport(subset,check_correlation=False)
.....
[...]

Issue #66, opened 2017-11-01
Warnings column names should be links to details in Variables section


Aylr:
## Benefits

- This would expedite navigation through large reports.
- I've seen a few click them with expectation of this behavior, so it seems natural.

Will submit a PR if you agree @JosPolfliet

romainx:
Hello,

If you have the time we will be happy to integrate this improvement in the new version.

Thanks !

conradoqg:
Hey,

With the above PR #82 merged, every variable name has a link for its respective analysis.

I think we can close this issue after those changes.

Best

romainx:
Hello, 

I confirm it's implemented now and very useful.

Thanks to all.

PR #82, opened 2018-01-02
Improve types handling

conradoqg:
The goal of this PR is to stabilize the handling of different types and to improve tests. The changes are:
- value_counts, nuniques and type inference are cached;
- boolean variances are treated correctly, eg. True, False, 0, 1 including NaN or not;
- types like List, Tuple, Dict are now officially unsupported until we improve them;
- mixed columns are also correctly handled;
- add correlation threshold option and check recoded correlation parameter (memory heavy)
- add variable anchor to improve navigation and small UX improvements (works both in stand-alone HTML and Jupyter notebook by using scrollTo instead of real link anchors)

This should solve #76 #77 #70 #44 #29 #66 

Please review my changes and if agreed merge it.

Best


romainx:
Great work!! 
I will have a closer look by the end of the day and merge it.
If fix I needed, we will do it later.

Thanks for your work

conradoqg:
Thanks for your time @romainx

conradoqg:
Hey, thanks for your observations!

> The unit test fails due to the cache used. That seems to not be cleared in some cases. It's better to call the clear cache at the beginning of describe. I'm not a big fan of caching in a global variable. But we can keep it as is. It's simple to remove if it causes problem.

I'm also not a fan of cache in global variables, indeed this part needs a good refactoring. I tried to simplify this first version before doing a refactoring. I evaluated some cache strategies like the lru-cache (built-in python 3) but it doesn't work since the function parameter is not a base type. (I also tried two other packages, but most of then relies on the parameter type)

Interesting is that the cache didn't fail in the tests here, not sure why.

> Using type as a variable name like in get_vartype is not a good practice since it can hide the build-in type function. We should use vartype instead

I agree.

> I will take into account new keyword arguments in the ProfileReport docstring. It's important since the help description will appear in Jupyter notebooks.

Ups, my mistake.

> I've not tried the unsupported type -> I will try to build a test case for it.
I added some test cases for the unsupported types, not sure if you mean to add a test case for that or add more tests.

I can do those changes if you need. Just ping me.

romainx:
Hello,

Thanks for the update I have just performed the changes you can review them. 
I may need your help to close all the corresponding issues:  #76 #77 #70 #44 #29 #66

I will have a look.

Many thanks

conradoqg:
OK, I'm responding each issue with the related fix/improvement done by this PR.

Best

conradoqg:
Done. I explained in each issue what was solved and requested the issue to be closed. In some issues I requested to you the openning of a new issue for a specific subject discussed in that issue (to separate the problem apart of new suggestions)

Best

Issue #40, opened 2017-01-23
Duplicate (and highly correlated) categoricals


dpavlic:
Duplicated categorical variables are not flagged:

![categorical_dupes](https://cloud.githubusercontent.com/assets/1075079/22222761/10a4c292-e186-11e6-8d84-294326cb20cd.png)

Perhaps it would be worthwhile to detect these types of exact duplicates, or maybe also perform a correlation analysis on categoricals?

tejaslodaya:
Yes, this has not been taken care of, because code uses `df.corr()` which is only applicable for numerical variables. For correlation of categorical variables, one should use Cramer's V.

Helper function -

```
def get_vartype(data):
    distinct_count=data.nunique(dropna=False)
    leng=len(data)
    if distinct_count <=1:
        return 'CONST'
    elif com.is_numeric_dtype(data):
        return 'NUM'
    elif com.is_datetime64_dtype(data):
        return 'DATE'
    elif distinct_count==leng:
        return 'UNIQUE'
    else:
        return 'CAT'
```

`import scipy.stats as stats`

```
def cramers_corrected_stat(confusion_matrix):
    chi2 = stats.chi2_contingency(confusion_matrix)[0]
    n = confusion_matrix.sum().sum()
    phi2 = chi2/n
    r,k = confusion_matrix.shape
    phi2corr = max(0, phi2 - ((k-1)*(r-1))/(n-1))
    rcorr = r - ((r-1)**2)/(n-1)
    kcorr = k - ((k-1)**2)/(n-1)
    return np.sqrt(phi2corr / min( (kcorr-1), (rcorr-1)))
```

Final call : 
```
for col1,s1 in df.iteritems():
    if get_vartype(s1)!='CAT':
        continue
    for col2,s2 in df.iteritems():

        if col1==col2:
            continue
        if get_vartype(s2)!='CAT':
            continue
        confusion_matrix=pd.crosstab(s1,s2)
        cramers_corr=cramers_corrected_stat(confusion_matrix)
        if(cramers_corr>0.9):
            print(cramers_corr,'\t',col1,'\t',col2)
```

CC: @dpavlic @JosPolfliet

JosPolfliet:
I implemented a draft for this, but in testing find that the Cramer's V is not specific enough to decide to reject a variable.
Example: a frequency matrix X/Y like this
```
X               Relict  Valid
Y 
Relict H                    1      0
Relict OC                  65      0
Relict iron                 1      0
[...]

PR #93, opened 2018-01-08
Implement suggestion #40

conradoqg:
This PR adds the following list of features to v1.5 solving the issue #40:

- Add cramers correlation for categorial variables

This is still a work in progress and I think we should say it's experimental.

I'll tell when it's done or mature enough to review and merge

Best


conradoqg:
The Cramers correlation suffers the same memory problem as the recoded correlation. I'm setting check_cramers to False by default like we did to check_correlation.

I think this implementation is done by now, please review and give me your opinion about this.

Best

PR #102, opened 2018-01-29
checking classes in correlations to avoid division by zero

amarczew:
when len(classes) is zero, an error is raised in correlation_matrix plot function

### DIFF
--- a/pandas_profiling/report.py
+++ b/pandas_profiling/report.py
@@ -192,9 +192,9 @@ def to_html(sample, stats_object):
 
     # Add plot of matrix correlation
     correlations = {}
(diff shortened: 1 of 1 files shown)


Aylr:
@amarczew Is there anything I can do to expedite this? Would you like some tests?

amarczew:
@Aylr I don't think so. I really had to change these lines to continue using it with my dataset when it does not have any correlations. A test for this could be using a dataset without correlations.

Aylr:
@JosPolfliet Is there anything we can do to expedite this being merged?

radomirbosak:
It would be great if we could merge this. The zero division error is quite annoying.

JosPolfliet:
is fixed in the meantime

PR #121, opened 2018-05-16
Handle max_freq == 0 in _format_row

ah-:
Fixes a minor issue where generating the profiling report would fail if `max_freq` ends up being 0. In that case `freq / max_freq` would be `NaN`, and converting `NaN` to int fails with an exception, breaking report generation.


ah-:
Thanks for merging!

Issue #10, opened 2016-04-22
Ploting a response variable on the histograms


Optimox:
Hey,

Great job with pandas-profiling I love it.
I think it would be great to have an extra parameter to specify a response column.
Plotting the average response for every bin of the histograms (for each variables) would allow to see obvious trends/correlations and would be useful for any regression problem (might be more tricky for classification where the response are discrete).
What do you think ?

Thanks!

JosPolfliet:
I have though about this too. For a classification target you can use a heat map too for example. It is a fair amount of work but could be very useful. Correlation for variables with the target, or even scatter plots with a smoother could be a nice addition. Feel free to implement some parts yourself and send a pull request.

JosPolfliet:
Not sure why I closed this, this should get the highest priority.

dartdog:
Nice, I'm plugging along on my much less elegant solution, finally got a few decent results, now need a basic refactoring to  see how generalized I can get. I can say the 1st few I have produced are hugely helpful...

dartdog:
F WWI Here is my very bad but maybe useful beginning on this, at least as I see it. I'll post again as I make progress, want to add the covariance calculation and then create the loop to cover the variables..

andyreagan:
I think this goes a long way toward adding this functionality:
https://github.com/JosPolfliet/pandas-profiling/pull/47

valtheval:
Hello, this is a nice feature.
Still, I'm faving problem using it. My pandas-profiling package is up to date but I have the following error when I type pandas_profiling.ProfileReport(df, compute_response=True)
`TypeError: _plot_histogram() got an unexpected keyword argument 'compute_response'`
even with my target variable named 'target'

Any help would be really appreciated,
Thanks

JosPolfliet:
Thanks guys, I have foreseen time this weekend to review this and other issues.

andyreagan:
This code is now too old to be smoothly integrated.

I've moved my contributions to a new repository at https://github.com/massmutual/robo-analyst

Optimox:
Hello guys,

I implemented this few month ago #124, I'll have quite some time in the next few weeks, could we agree on what we would like to see in this PR in order to merge it?

I'd be happy to help! Let me know! 
Cheers!

Optimox:
@JosPolfliet @sbrugman Any update here?

sbrugman:
Hi @Optimox,

I'll give you some context, then we can discuss the best steps forward.

[...]

PR #124, opened 2018-06-04
Plots with target issue#10

Optimox:
pdpf_target_example
This code implements the possibility of specifying a target variable when launching a ProfileReport only by specifying target_name = "the_target_column_name". The average value of this target is then plotted on every bins of the histograms. This can help to recognize obvious patterns and correlations only by looking at pandas-profilling results.

This implementation is what I had in mind when I created the issue #10 a long time ago.
I tried to made the changes as minimal as possible. I hope this could finally close this issue!

I also made some changes so that this version is python2 friendly by adding some float() in the current code.

There is still room for improvement in the final visualization but I feel like this could still be very useful for users without any extra cost (if you don't specify a target nothing changes). Let me known if you think that there are improvements that I could do in order to close issue #10.

Thanks!

### DIFF
--- a/pandas_profiling/__init__.py
+++ b/pandas_profiling/__init__.py
@@ -61,6 +61,10 @@ class ProfileReport(object):
     def __init__(self, df, **kwargs):
         """Constructor see class documentation
         """
+
+        if ("target_name" in kwargs.keys()):
+            kwargs['target_series'] = df[kwargs['target_name']]
+
         sample = kwargs.get('sample', df.head())
 
         description_set = describe_df(df, **kwargs)
@@ -101,7 +105,7 @@ class ProfileReport(object):
         result = []
         if hasattr(variable_profile, 'correlation'):
             result = variable_profile.index[variable_profile.correlation > threshold].tolist()
-        return  result
+        return result
 
     def to_file(self, outputfile=DEFAULT_OUTPUTFILE):
         """Write the report to a file.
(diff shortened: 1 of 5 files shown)


sl1102:
Hi, I tried aht_profile=pandas_profiling.ProfileReport(df_pd,  target_name = 'handle_time') but it gave me an error message saying 'TypeError: _plot_histogram() got an unexpected keyword argument 'target_name''.  Is this the right way to use it? Thank you.

Optimox:
Hi, yep that's how it's supposed to be used, this error looks weird, I just tried it again on my computer on Titanic dataset and it works like a charm. Are you sure to be using this branch when importing pandas profiling? 
Since it's not merged (yet) you need to be sure that you installed this particular version, that's the first thing that came to mind when seeing your error, but with more info maybe I could help you more.

Optimox:
I mean that you need to install this specific version of pandas profiling from my personal repo : 
- uninstall your current version of pandas profiling with `pip uninstall pandas-profiling`
-  clone my repo `git clone git@github.com:Optimox/pandas-profiling.git`
- then go to pandas-profiling directory and run `pip install .` and you should be able to use this version

sargupta:
This seems interesting and useful to me but when I tried to plot got the error (even after cloning as recommended above):
`_plot_histogram() got an unexpected keyword argument 'target_name'`

Optimox:
hi @sargupta,

I might have forgotten to specify that you should go to the correct branch of my implementation.

Here are clearer step that should work just fine:
- uninstall your current version of pandas profiling with `pip uninstall pandas-profiling`
- clone my repo:  `git clone git@github.com:Optimox/pandas-profiling.git`
- go to the directory: ` cd ./pandas-profiling`
- checkout to the target branch : `git checkout plots_with_target`
- then install pandas in editable mode so that you can switch from one branch to an other : `pip install -e .`

Please let me know if it still does not work. I'll try to rebase my fork so that it is up to date with new changes.

PR #139, opened 2018-09-20
fix recoded test with optional check_recoded flag

sleitner:
This is needed for the CategoricalDataTest to pass.


sleitner:
sorry for the crud -- opening from a different branch

PR #141, opened 2018-09-24
fix recoded test with optional check_recoded flag

sleitner:
This is needed for the CategoricalDataTest to pass.


sleitner:
@JosPolfliet can you say something about why you're closing?

JosPolfliet:
I fixed the same bug while working on another pull request :). Thanks for
your contribution!

PR #144, opened 2018-11-06
Fixed wrong NaT hist bins in plot.py

dsvolk:
A simple fix to the problem when NaT values in a dataframe result in having histogram bins for year 1680. 
This makes the histogram very uninformative. 
After the fix, only proper datetime values contribute to all the time histograms.


romainx:
Thanks for the fix.

PR #146, opened 2018-11-20
updated README example

jameslamb:
Hello,

Thank you for this awesome project! I was working through the docs today, and hit something in the README. The examples use `pd.read_csv()` to read in a CSV, but as far as I can tell this is not necessary to provide an example for users to try out `pandas-profiling`.

I hope you will consider this PR to change the README so the examples can be run just by copying and pasting the code. I think this will lead to a slightly better user experience for first-time visitors to the project.

-James


romainx:
Hello,

You're right. It's better like this. Thank you for your time.

Regards

PR #175, opened 2019-06-19
Add support for pyinstaller

elias-j-neuman:
Add documentation on README on how to package into pyinstaller application

Tweaked how the templates are loaded and how the matplotlib style guide is selected

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,25 @@ A set of options are available in order to adapt the report generated.
     * `check_recoded` (`boolean`): Whether or not to check recoded correlation (`False` by default). Since it's an expensive computation it can be activated for small datasets.
 * `pool_size` (`int`): Number of workers in thread pool. The default is equal to the number of CPU.
 
+## Package with pyinstaller
+
+To use pandas-profiling within a pyinstaller packaged application, you need to follow these steps:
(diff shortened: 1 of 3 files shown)


sbrugman:
In what scenario would someone use pyinstaller in combination with `pandas-profiling`? What are the main advantages (and disadvantages)? Should be added to the README in case we want to include this.

kirancrazy99:
Hi 

It would be great if pandas-profiling support pyinstaller, as I have tried many times I was unable to create a exe file with pyinstaller only when importin pandas-profiling in my script, if I remove importing pandas package then exe file getting created..

This is my requirement in current project  that I have to  create a Data Profiling app that any user can use even with out having install python ..

please guide me how it make exe by importing pandas-profiling package as well

Platform:
Python 3:7
pandas profiling 
pyinsaller
tkinter
windows 64bit
ram 16gb

bruce0715:
Really need the guides about using pandas-profiling under pyinstaller. it tells me cannot find config_default.yaml

bruce0715:
> Add documentation on README on how to package into pyinstaller application
> 
> Tweaked how the templates are loaded and how the matplotlib style guide is selected

This doesn't work~

PR #189, opened 2019-06-23
Install using Pipenv from githug repo or pypi

Sim4n6:


sbrugman:
Does this add information that users do not know? 

Users of `pipenv` should be able to translate `pip install (...)` into `pipenv install (...)` in my opinion.

Issue #244, opened 2019-08-20
progress bar is needed


appleyuchi:
**Is your feature request related to a problem? Please describe.**
when the datasets is very large ,then data.profile_report wil be stuck with no response,
Of course I understand it is busy,but ....
I really want a progress bar....

**Describe the solution you'd like**
I really want a progress bar....

**In which situation?**
in kaggle "IEEE-CIS Fraud Detection",the link is:
https://www.kaggle.com/c/ieee-fraud-detection/data

kaggle api will help you download the datasets very quickly if your internet is slow sometimes.

**Describe alternatives you've considered**
seems no way now

**Additional context**

Thanks for your work on this package.

marco-cardoso:
Hello guys, I'd like to help building this functionality, but I'm new to the open source world. I already read the contributing guide but I've not found technical details. Do I just need to create a new branch and start coding on it, and after is completed do a merge request with master ?

Thanks

sbrugman:
@marco-cardoso Yes, that's about it. There is room for changes in the pull request, so you don't have to worry too much about getting it perfect the first time. You might also want to add a small paragraph in the contribution guide, where you would have expected the technical details for future reference.

alfonsomhc:
+1

TylerWakefield:
+1

ankushkundaliya:
@marco-cardoso I would like to collaborate with you on this. Please let me know how may I help in adding this functionality. I am also new to the open-source. Thanks.

jvschoen:
I tried using the new feature in v2.4.1 in jupyter notebook and I just get this printed out:

HBox(children=(FloatProgress(value=0.0, description='variables', max=41.0, style=ProgressStyle(description_wid…

HBox(children=(FloatProgress(value=0.0, description='correlations', max=1.0, style=ProgressStyle(description_w…

HBox(children=(FloatProgress(value=0.0, description='table', max=1.0, style=ProgressStyle(description_width='i…

HBox(children=(FloatProgress(value=0.0, description='warnings', max=3.0, style=ProgressStyle(description_width…

HBox(children=(FloatProgress(value=0.0, description='package', max=1.0, style=ProgressStyle(description_width=…

HBox(children=(FloatProgress(value=0.0, description='build report structure', max=1.0, style=ProgressStyle(des…

I am using the most recent version on github and using the config_minimal file

sbrugman:
@jvschoen https://stackoverflow.com/a/57343401/470433

[...]

PR #255, opened 2019-09-16
Simple progress bar

marco-cardoso:
#244  This version of the progress bar should work in a normal CLI and also in jupyter notebook.  I tried to figure out a way to generate a more realistic update status, but for the sake of simplicity I used fixed percentage values. If you know about a better way to do that, please let me know.


pandas_profiling/utils/notebook.py
```diff
@@ -6,3 +6,13 @@ def full_width():
     from IPython.core.display import display, HTML
 
     display(HTML("<style>.container { width:100% !important; }</style>"))
+
+
+def in_ipynb() -> bool:
```
marco-cardoso: I've not found a specific method to do this in the another modules, so I decided to add it. Let me know if there's a better approach to detect the environment.

sbrugman: > It is also possible to let tqdm automatically choose between console or notebook versions by using the autonotebook submodule:
> ```from tqdm.autonotebook import tqdm```

https://github.com/tqdm/tqdm

sbrugman:
Great work, @marco-cardoso. 

Some points to complete the contribution:
-  `tdqm` itself contains a mechanism to differntiate between environments (see above). 
- It should be possible to disable the progressbar (`tqmd(disable=...)`).
- There are failing tests (`ImportError: IntProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html`).
- Please run the code linting and regenerate the docs (`make lint` and `make docs`, more [details](https://github.com/pandas-profiling/pandas-profiling/blob/master/Makefile)).

marco-cardoso:
Thanks for the points @sbrugman . I did a couple of changes and now the codes are working and the documents are updated. The default option is not to show the progress bar, in order to activate it is necessary to set the parameter progress_bar= True when calling the method profile_report.

appleyuchi:
But I did not see any change in my kaggle notebook
Which edition of profiling is progress bar  supported with?

appleyuchi:
Config parameter "progress_bar" does not exist.

marco-cardoso:
Hello @appleyuchi , thanks for your feedback.

Could you tell me which version did you use?

I tested the version from my account, available [here](https://github.com/marco-cardoso/pandas-profiling/tree/progress-bar) (since this PR was not approved yet), and after executing 'python setup.py install' and importing pandas_profiling in a script, an error was raised because TQDM was not installed. After I fixed it, by installing with pip separated, the progress bar worked as expected. This made me realize that I forgot to update the setup.py with the progress bar dependencies (Only updated the requirements.txt). 

Also, I recorded a short video with the progress bar working with a small dataset.

ezgif-2-e392bd305f71

marco-cardoso:
I'm trying to update this, but something is wrong. Do I need to write unit tests for the methods which I created in order for the PR to pass the coverage tests?

appleyuchi:
> Hello @appleyuchi , thanks for your feedback.
> 
> Could you tell me which version did you use?
> 
> I tested the version from my account, available [here](https://github.com/marco-cardoso/pandas-profiling/tree/progress-bar) (since this PR was not approved yet), and after executing 'python setup.py install' and importing pandas_profiling in a script, an error was raised because TQDM was not installed. After I fixed it, by installing with pip separated, the progress bar worked as expected. This made me realize that I forgot to update the setup.py with the progress bar dependencies (I had only updated the requirements.txt).
> 
> Also, I recorded a short video with the progress bar working with a small dataset.
> 
> ezgif-2-e392bd305f71

Thanks for your replies~!
My Environment is as follows:
Python version:3.6.7rc1
tqdm                   4.34.0
pandas                 0.23.4   
pandas-profiling       2.3.0

sbrugman:
@marco-cardoso Many thanks for kickstarting this feature. A lot has changed since the PR, so it seemed to me to be more effective to implement it in a new PR than resolve the merge conflicts (#345).

PR #276, opened 2019-10-16
feat: add option to overwrite default with custom config file

lvwerra:
This feature is useful for using a custom config file to overwrite all or parts of the default configuration without the need to change the code and pass the settings as kwargs.


lvwerra:
Hey, do you have any feedback on this PR? Anything I should add such that it could be merged?

PR #289, opened 2019-11-25
Adding docs on how to install pandas-profiling from inside the Jupyter notebook

neomatrix369:
Over the weekend I was at a hackathon and I shared with others about pandas-profiling.

We used it to assess a number of datasets we had to work on. It was pretty useful and helpful to get a quick insight into the dataset.

For Windows, I learnt from one of the teammates that we could install pandas-profiling from inside the notebook using a line as simple as:

```bash
import sys
!{sys.executable} -m pip install pandas-profiling
```

And in principle, we could do the same for installation via `conda` as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,6 @@ You can install using the pip package manager by running
 Alternatively, you could install directly from Github:
 
     pip install https://github.com/pandas-profiling/pandas-profiling/archive/master.zip
-
     
 ### Using conda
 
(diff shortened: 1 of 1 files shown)


neomatrix369:
@sbrugman Happy New Year! Great to see you active again. 

Would like me to resubmit this PR with the merge conflicts resolved? I saw you closed it without any comment, just curious if such contributions are desired and welcome.

sbrugman:
@neomatrix369 I closed the PR for two reasons. The information might be somewhat misleading for the reason why we want to use `!{sys.executable}` (see [this post](https://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/#How-to-use-Pip-from-the-Jupyter-Notebook)). I also feel that the information does not belong on the main documentation, as users of notebooks should know this to use _any_ package.

I hope that closing the PR does not discourage you from contributing further. Your input is valued. Even without being on the main page, this information can serve as reference material for some people in the future.

neomatrix369:
> @neomatrix369 I closed the PR for two reasons. The information might be somewhat misleading for the reason why we want to use `!{sys.executable}` (see [this post](https://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/#How-to-use-Pip-from-the-Jupyter-Notebook)). I also feel that the information does not belong on the main documentation, as users of notebooks should know this to use _any_ package.
> 
> I hope that closing the PR does not discourage you from contributing further. Your input is valued. Even without being on the main page, this information can serve as reference material for some people in the future.

Thanks for coming back to me with a response. Firstly I'll beg to differ about everyone knowing how to do something via notebooks - it's quite the opposite in my experience. Even myself I'm learning how to do sophisticated things. And especially it's a bit challenging when it comes to Notebooks on the Windows platform (which applies in this case). Where would you have liked such information? I think the automated docs folder is a good place to look to add such documentation - would you agree?

Thanks for encouraging me to contribute, I'll take up on that the next time I see another opportunity.

PR #299, opened 2019-12-27
Rejected Columns based on co-relation type

Enforcer007:
This PR is raised against to the Feature Request raised: https://github.com/pandas-profiling/pandas-profiling/issues/298

Helps in selecting type of Co-relation to be used in rejected Columns.

'pearson', 'spearman', 'kendall', 'phi_k'


sbrugman:
Great addition, thank you. I'll add it when we've resolved the dependency issues.

PR #300, opened 2019-12-29
Require numpy minimum version

gliptak:
https://travis-ci.com/pandas-profiling/pandas-profiling/jobs/270608894
```astropy 4.0 has requirement numpy>=1.16, but you'll have numpy 1.15.4 which is incompatible.```


sbrugman:
Thank you, merging it together with #296.

PR #397, opened 2020-03-03
Update describe.py to fix n_infinite

WesleyTheGeolien:
Hey guys I was reading through the code and came across this:
```
count = series.count()
n_inifinte = count - series.count()
```

So I am pretty sure n_infinite is always 0. I believe the following expression is correct:

```
leng = len(series)
count = series.count()
n_inifinite = leng - count
```

Hope this helps


sbrugman:
Thanks for reporting this, greatly appreciated. This is somewhat legacy code.

The tests are failing now. In at least one case (test_describe.py) it recognizes `None` as an infinite value. There is no test case containing infinite values, we should add one. 

We can merge the PR once the tests are updated, would you have time for that?

sbrugman:
Superseded by #432.

Issue #279, opened 2019-10-25
Compatibility with Dask.DataFrame


endu:
I want to use pandas-profiling with Dask.DataFrame.

Now, when I use Dask.DataFrame, pandas-profiling shows error message.

```
dd_data = dd.read_csv('data.csv',assume_missing=True)
pr_data = pp.ProfileReport(dd_data)
:
/usr/lib/python3.6/site-packages/pandas_profiling/utils/dataframe.py in rename_index(df)
     93         The DataFrame with {index,column} `index` replaced by `df_index`, unchanged if the DataFrame does not contain such a string.
     94     """
---> 95     df.rename(columns={"index": "df_index"}, inplace=True)
     96 
     97     if "index" in df.index.names:

TypeError: rename() got an unexpected keyword argument 'inplace'
```

Is it difficult to have compatibility with Dask.DataFrame ?

sbrugman:
Making `pandas-profiling` compatible with the Dask.DataFrame would be an enormeous operation. My opinion is that the package shouldn't introduce dependencies by doing too much.

However, luckily @devin-petersohn is working on precisely that. We are eagerly waiting for it's stable release. If you haven't already, check out the project here:

https://github.com/modin-project/modin

endu:
I've acknowledged. And, I did not know modin-project. That looks very nice!
Thanks for your response.

brunompacheco:
I've developed some work on modifying pandas-profiling to make it work for dask dataframes and successfully generated some reports.

I've developed it in the company I work for (therefore, the code is currently on our private repo), but I already received the green light to publish the code, just need to mention them. Also, since this stayed in the low-priority side of my tasks for a long time, the fork is from version 1.4.2, but I'm looking forward to update it to the current version.

By the way, should I move this to #227?

sbrugman:
@brunompacheco Great to hear. My recommendation would be to start a PR and continue the discussion from there.

brunompacheco:
@sbrugman Ok, I'll move the code to GitHub asap.

PR #403, opened 2020-03-05
Add support for Dask DataFrames

brunompacheco:
As I said in #279, this is my work so far. I changed the name of the lib so I could have both packages installed.

As you can see, it's far from a finished work for:

1. It's from an old version
1. I replaced the pandas functions, therefore it is concurrent to the original
1. Many features were lost in the process

But I'm willing to put the effort into it, so I created this mostly to discuss what would be a great approach for adding support for dask dataframes.


sbrugman:
@brunompacheco Thanks for this exciting contribution. Performance is one of the biggest downsides of the current implementation of pandas-profiling. Dask would be a great candidate to scale up. 

For the reasons you mention, working further in this direction requires dedicated effort to produce a dak. Moreover, maintaining two code bases for two in parallel integrations for pandas and Dask introduces significant technical debt. We do want to support multiple integrations, but supporting them are not a core task of pandas-profiling. [Modin](https://github.com/modin-project/modin) seems to most suitable candidate for this at this moment.

For now I would like park this PR for a while, we might come back to it later.

Issue #389, opened 2020-02-25
Data have rejected variables, but do not display in Warnings


NickYi1990:
**Data have rejected variables, but do not display in Warnings**

```python
from pandas_profiling import ProfileReport
profile = ProfileReport(pd.DataFrame({"a": [1,1,1,1,1,1]}), minimal=True)
```
![image](https://user-images.githubusercontent.com/13381684/75269098-56240900-5833-11ea-8774-b588d3ec874c.png)

**2.5.0**

sbrugman:
Thank you for reporting this. We invite interested contributors to create a pull request with this functionality. More details can be found in the [contribution guide](https://github.com/pandas-profiling/pandas-profiling/blob/master/CONTRIBUTING.md).

loopyme:
I noticed that four types of warning messages(`Uniform`,`Unique`,`Rejected`,`Constant`) are ignored when rendering. Since `Rejected` is always comes with other warnings, is it necessary to add all types of warnings above?

or it will end up like this:
![Screenshot-20200413140323-1283x574](https://user-images.githubusercontent.com/33747091/79096552-19cf4c80-7d90-11ea-9693-068790bab0c2.png)

sbrugman:
v2.7.0 changes the behavior to report all warnings, except "rejected", as that would be redundant.

PR #433, opened 2020-04-13
Display all types of messages in Warnings.

loopyme:
Working on #389

**About CI**: I run a trigger build, which failed while running test case:`tests/unit/test_example.py`. It is caused by `assert \"<span class=badge>10</span>\" in profile.to_html()` while the total number of warnings is 10 and all types of warnings are shown in my branch. 

Any ideas? Should I update the test case or abort this new functionality?

### DIFF
--- a/src/pandas_profiling/report/presentation/flavours/html/templates/warnings/warning_uniform.html
+++ b/src/pandas_profiling/report/presentation/flavours/html/templates/warnings/warning_uniform.html
@@ -3,5 +3,5 @@
     distributed
 </td>
 <td>
-    <span class="label label-warning">Constant</span>
+    <span class="label label-info">Uniform</span>
 </td>
--- a/src/pandas_profiling/report/presentation/flavours/html/templates/warnings/warning_uniform.html
+++ b/src/pandas_profiling/report/presentation/flavours/html/templates/warnings/warning_uniform.html
@@ -1,0 +1,7 @@
+<td>
+    <a class="anchor" href="#pp_var_{{ message.anchor_id }}"><code>{{ message.column_name }}</code></a> is uniformly
+    distributed
+</td>
+<td>
+    <span class="label label-warning">Constant</span>
+</td>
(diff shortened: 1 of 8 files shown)


src/pandas_profiling/report/presentation/flavours/html/templates/warnings/warning_uniform.html
```diff
@@ -0,0 +1,7 @@
+<td>
+    <a class="anchor" href="#pp_var_{{ message.anchor_id }}"><code>{{ message.column_name }}</code></a> is uniformly
+    distributed
+</td>
+<td>
+    <span class="label label-warning">Constant</span>
```
sbrugman, line 6: Constant = Uniform

src/pandas_profiling/report/presentation/flavours/html/templates/warnings/warning_unique.html
```diff
@@ -0,0 +1,7 @@
+<td>
+    <a class="anchor" href="#pp_var_{{ message.anchor_id }}"><code>{{ message.column_name }}</code></a> has unique
+    values
+</td>
+<td>
+    <span class="label label-warning">Constant</span>
```
sbrugman, line 6: Constant => Unique

sbrugman:
> Working on #389
> 
> **About CI**: I run a trigger build, which failed while running test case:`tests/unit/test_example.py`. It is caused by `assert \"<span class=badge>10</span>\" in profile.to_html()` while the total number of warnings is 10 and all types of warnings are shown in my branch.
> 
> Any ideas? Should I update the test case or abort this new functionality?

You are right that this is causing the issue. The test should be updated, as the expected behaviour has changed: we now expect > 10 warnings.

loopyme:
Sorry about the mistakes I make, I have fix them in my latest commit.

Besides, I keep ignoring the `Rejected` warnings, since `Rejected`=(`Unsupported` or `Constant`), display both of them seems redundant.

Issue #434, opened 2020-04-14
Profiling stuck on variables reaching 99%


zkid18:
**Describe the bug**

The profiling being stuck on variables stage reaching 99%.
Been waiting for about 2 hours without any further progress.
This error occurs when I sample my data while it working fine on the initial dataset
Any hints how I can debug this issue?

variables:  99%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ | 103/104 [22:26<03:12, 192.91s/it]

**To Reproduce**
_Data:_
```
Int64Index: 860029 entries, 2264008 to 2388942
Columns: 103 entries, access_date to log_date
dtypes: Int64(69), float64(5), object(29)
memory usage: 739.0+ MB
None
2020.04.14 16:41:35] I Start profiling for df with shape (860 029, 103)
```

```
def create_eda_html(df, report_file_name, is_sampled):
     df_to_profile = df
     if is_sampled:
         split = StratifiedShuffleSplit(n_splits=1, test_size=0.2, random_state=42)
         for _, sampled_index in split.split(df, df['label']):
             strat_df = df.loc[sampled_index]
         df_to_profile = strat_df
     logging.info("DF info \n", df_profile.info())
     logging.info("Start profiling for df with shape {}".format(df_to_profile.shape))
     profile = ProfileReport(df_to_profile, title='Pandas Profiling Report', html={'style': {'full_width': True}})
     if os.path.exists(os.path.join(ROOT_DIR, report_file_name)) and report_file_name.endswith('.html'):
         profile.to_file(output_file=os.path.join(ROOT_DIR, report_file_name))
     else:
         profile.to_file(output_file="your_report.html")
     logging.info("Pandas profiling file created")
```

**Version information:**

Version information is essential in reproducing and resolving bugs. Please report:

* _Python version_: Python 3.6.7
* _Environment_:  cmd
* _`pip`_: 

<details><summary>Click to expand <strong><em></em></strong></summary>
<p>

```
appnope==0.1.0
astroid==2.3.3
astropy==4.0.1.post1
attr==0.3.1
attrs==19.3.0
backcall==0.1.0
beautifulsoup4==4.8.2
bleach==3.1.4
bs4==0.0.1
certifi==2019.11.28
chardet==3.0.4
confuse==1.0.0
cycler==0.10.0
decorator==4.4.2
defusedxml==0.6.0
entrypoints==0.3
[...]

Issue #421, opened 2020-03-27
Incremental building or build in phases feature


neomatrix369:
**Missing functionality**

It's a bit annoying when after a long profiling task the process fails due to one of these:
- output type not specified (if output file is missing an extension), not generating output report for you 
- Failed to allocate an array of 2.09GB
- or any other reasons that it finds out only later on

Incremental building is an efficient way to improve the performance and efficiency of systems.

Basically I'm suggesting a pipeline-like approach where one stage feeds into the next, so they can be built separately and started and finished separately. I came up with this idea when I had to re-run the task the third time cause something I missed out or made a mistake caused it to break the flow.

The other benefit of this is, to be able to create reports up to the phase we have built that way we can pick and choose the phases/reports we like to have.

**Proposed feature**
If PP caches the output of each phase of the process somewhere such that even if we broke the process, we can continue from where we left and/or if we want to build in stages because of various reasons then it would help a lot.

Especially if it is a trivial mistake and change the parameter and rerunning I can fix the workflow.

**Alternatives considered**

None so far. Maybe refactor the app to create the pipeline feature.
It would also mean to break the big parts of the system into smaller ones and keep track of the I/O dependency tree, so we know which phase/piece of the system needs to be built first for the next ones to follow through.

loopyme:
Hi, I am a big fan of PP and recently finished reading most of the codes.

I have an idea that seems good about pipelining the project. The main idea is to create an option for user to incremental building.

To maintain current software behavior, we can just add a boolean parameter at `ProfileReport.__init__`(Let's say `lazy`). When it set to True, init function won't run `describe_df` or build the report, but store the df. The `describe_df` will be called when `self.report` is accessed.

It will implement like this:
```python
@property
def report(self):
    if self._report is None:
        # ... compute self._report
    return self._report
```

What's more, I think PP will be more scalable if we split the `ProfileReport.__init__` in to some functions and add functions to load and dump `self.report`(Since it always take most of the time).

The improvement will maintain current software behavior, make it possible for user to arrange the workflow, and may reduce the running time especially when user doesn't want a html report.

I'll working on the PR if this idea is acceptable and agreed to be effective.

sbrugman:
Sure, go ahead!

neomatrix369:
> Hi, I am a big fan of PP and recently finished reading most of the codes.
> 
> I have an idea that seems good about pipelining the project. The main idea is to create an option for user to incremental building.
> 
[...]

Issue #420, opened 2020-03-27
[Question] What are the different options we can use when running analysis on a big/large datasets?


neomatrix369:
**Question**
It's a known query that when analysing large datasets we have to wait for a long time. Although there is some work in the background to improve performance. Is this going to use Python's internal/external libraries for multiprocessing, concurrency? 

How do you make PP use multiple CPUs atm or have people been able to run it using frameworks like Dask, Swifter and the likes?

I tried `modin` with PP the last time and there were type compatibility issues so I reverted to not using it.

But then there are parameters we can pass in to make it not do expensive computation. Is there a middle-ground where you wait for less but get a better report, that way we can incrementally move up depending on the data each profiling produces (I hope you get my drift).

**Proposed feature**
For the short run, examples in the docs/README with different parameters that can be passed in when profiling large datasets. What I really mean by this is:

- If you want _something quick_, but happy with _a brief report_, use these parameters (and show examples)
- For a half-way balance between _wait-time and depth of the reports_, use these parameters (and show examples)
- If you are _okay to wait for long, but then want a detailed report_, use these parameters (and show examples)

Also, examples of how PP can be used to scale up using external tools/frameworks/libraries will be of great help. There are discussions about this in the Issues and PRs but not seen them back in the Docs or Readme.

**Alternatives considered**
Go through the issues and Docs. 

Git issues to look into for hints/examples:
- https://github.com/pandas-profiling/pandas-profiling/issues/222
- https://github.com/pandas-profiling/pandas-profiling/issues/76
- https://github.com/pandas-profiling/pandas-profiling/issues/30
- https://github.com/pandas-profiling/pandas-profiling/issues/279
- https://github.com/pandas-profiling/pandas-profiling/issues/293
- https://github.com/pandas-profiling/pandas-profiling/issues/258
- https://github.com/pandas-profiling/pandas-profiling/issues/245
- https://github.com/pandas-profiling/pandas-profiling/issues/227

Docs
- https://github.com/pandas-profiling/pandas-profiling#documentation
- https://pandas-profiling.github.io/pandas-profiling/docs/

neomatrix369:
I think I might have come across something that might help as much as examples would:
- [default configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_default.yaml)
- [minimal configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_minimal.yaml)
[...]

PR #435, opened 2020-04-17
Some preparation works for pipeline

loopyme:
Lots of code are changed, I have already double checked the code and run some tests. I would much appreciate if someone can spend some time reviewing it.

## What I've done
- Make ProfileReport lazy by default (Compute when need)
   - Light weight of series_descriptions (Already passed some tests, but I'm not sure if there's any potential harm)
   - Add dump and load for ProfileReport (Smaller and more comprehensive than json)
 - Refactor the describe function
   - Refactor and split some related functions to make the stages and data flow clear
   - Use a unified progress bar
   - Add dask-based compute graph solution: ddescribe (Experimental but might be faster)
   - Add tests for ddescribe

## Changed behaviors
 - Progress bar number reduced: describe with a unified progress bar
 - Default to be lazy, so object instantiating does not trigger full computation.
 - Add warning messages for 'Constant', 'Uniform' and 'Unique'

## New behaviors
 - Add an option: 'use_dask' (Default:False)
 - Add method dump and load for `.ProfileReport`

## Related PR
 - #389 : Add warning messages for 'Constant', 'Uniform' and 'Unique' 
 - #420 : Add an option:'use_dask'
 - #421 : Pipeline `.model.describe`
 - #434 : Display feature name when describe

## Details
 - Refactor `.ProfileReport.__init__`
   - Strengthen `.config`
     - Make it able to update, compare and check if it's default
     - Add function to restore config to default
     - Warning when incoming parameter appears to be the default but config is not
   - If lazy, `.ProfileReport.description_set` and `.ProfileReport.report` will compute when need
   - Make `.ProfileReport` able to save and load
 - Refactor `.model.describe`
   - Pipeline the `.model.describe.describe` method
   - Add `.model.describe.ddescribe` method, which implement `.model.describe.describe` with dask-based compute graph.
   - Add 'use_dask' option to default yamls. (False by default) 
   - Move some methods from `.model.describe` to `.model.statistics`
   - Split `.model.correlation.calculate_correlation` in to smaller pieces
 - Add warning messages for 'Constant', 'Uniform' and 'Unique'
 - Rearrange the progress bars
   - Move report_structure bar into `.report.structure.report.get_report_structure` function
   - Run `.model.describe.describe` with a unified progress bar
 - Optimise imports
 - Update tests
   - In `tests.notebooks.meteorites`, change progress bars count to 2.
   - In `tests.notebooks.titanic`, change progress bars count to 2.
   - In `tests.unit.test_example`, change warning message count to 12.
   - In `tests.unit.test_describe`, add tests for `.model.describe.ddescribe`.
   - Add `tests.unit.test_load`, which test `.ProfileReport.dump` and `.ProfileReport.load`
 - Update requirements-test: Dask is considered as an optional dependency.


sbrugman:
@loopyme Great contribution, especially considering it's your first time. I'd be happy to merge them when the code review has finished.

Before we start fully reviewing the code:
- The docs are now the predominant part of the file changes. This convolutes the code review. Could you revert these changes? Our new Github Actions script will take care of it later.
- This PR includes multiple features, which in principle is great, if all features are equally stable and will be in the next release. Profiling using Dask seems to lack some features compared to the original version. Preferably, we move less stable features to a new branch.

loopyme:
Sure, I have already reverted all changes about dask and doc in current branch. 

Should I make a PR from my dask branch to another branch here?

loopyme:
BTW, The "dask branch" only uses dask to schedule the computation steps in the description method. I have ran some tests, so I considered it as an optional experimental feature.

src/pandas_profiling/__init__.py
```diff
@@ -98,9 +140,31 @@ def get_sample(self, df: pd.DataFrame) -> dict:
...
+        if self._description_set is None:
+            self._description_set = describe_df(self.df)
+            self.date_end = datetime.utcnow()
+        return self._description_set
+
+    @property
+    def report(self):
+        if self._report is None:
+            self._report = get_report_structure(
+                self.date_start, self.date_end, self.sample, self.description_set
```
sbrugman, line 161: `self.date_end` is not set since `self.description_set` is not loaded.

src/pandas_profiling/__init__.py
```diff
@@ -4,48 +4,107 @@
...
             )
 
-        if minimal:
-            config_file = get_config_minimal()
+        if df is None and not lazy:
+            raise ValueError("Can init a not-lazy ProfileReport with no DataFrame")
+
+        if not config_file and not minimal and not config.is_default:
+            warnings.warn(
+                "The configuration of pandas profiling is universal, which will affect all ProfileReport "
```
sbrugman, line 54: What is the function of this warning?

src/pandas_profiling/__init__.py
```diff
@@ -4,48 +4,107 @@
 """
 import json
 import warnings
-from pathlib import Path
 from datetime import datetime
+from pathlib import Path
+from typing import Union
 
-import pandas as pd
+import joblib
```
sbrugman, line 11: It seems this dependency is only used to hash the DataFrame. Is this really necessary?

These might be relevant:
http://drorata.github.io/posts/2017/May/26/when-trying-to-hash-a-data-frame/index.html
https://github.com/pandas-dev/pandas/issues/16517#issuecomment-304364225

src/pandas_profiling/__init__.py
```diff
@@ -118,9 +182,9 @@ def get_rejected_variables(self) -> set:
             if message.message_type == MessageType.REJECTED
         }
 
-    def to_file(self, output_file: Path, silent: bool = True) -> None:
+    def to_file(self, output_file: Union[Path, str], silent: bool = True) -> None:
         """Write the report to a file.
-        
+
         By default a name is generated.
```
sbrugman, line 188: The docstring is out-of-date. Please include .pp. You might also want to remove the out-of-date "By default a name is generated."

src/pandas_profiling/__init__.py
```diff
@@ -271,3 +330,146 @@ def to_app(self):
...
+        return ""
+
+    def dumps(self) -> bytes:
+        """
+        Serialize ProfileReport and return bytes for reproducing ProfileReport or Caching.
+
+        Returns:
+            Bytes which contains hash of DataFrame, config, sample, _description_set and _report
+        """
+        import pickle
```
sbrugman, line 349: Why choose pickle? Have you considered faster alternatives?

https://www.benfrederickson.com/dont-pickle-your-data/

src/pandas_profiling/__init__.py
```diff
@@ -271,3 +330,146 @@ def to_app(self):
...
+
+    def _repr_html_(self):
+        """The ipython notebook widgets user interface gets called by the jupyter notebook."""
+        self.to_notebook_iframe()
+
+    def __repr__(self):
+        """Override so that Jupyter Notebook does not print the object."""
+        return ""
+
+    def dumps(self) -> bytes:
```
sbrugman, line 342: Style suggestion: implement the serialization features as a mixin class. 

Example for JSON:
https://gist.github.com/jimbattin/dd5cac131b582d8201d18992d52b6fdb

src/pandas_profiling/model/correlations.py
```diff
@@ -135,130 +134,94 @@ def warn_correlation(correlation_name, error):
     )
 
 
-def calculate_correlations(df: pd.DataFrame, variables: dict) -> dict:
+def calculate_correlation(df: pd.DataFrame, variables: dict, correlation_name: str):
```
sbrugman, line 137: Add return type

src/pandas_profiling/model/describe.py
```diff
@@ -571,87 +63,80 @@ def describe(df: pd.DataFrame) -> dict:
...
+            "cramers",
+            "recoded",
+        ]
+        if config["correlations"][correlation_name]["calculate"].get(bool)
+    ]
+
+    number_of_task = 7 + len(df.columns) + len(correlation_names)
+
+    with tqdm(
+        total=number_of_task, desc="Describe", disable=disable_progress_bar
```
sbrugman, line 82: "Describe" might not be the most informative description. How about "Profiling"?

src/pandas_profiling/model/statistic.py
```diff
@@ -0,0 +1,591 @@
+"""Compute statistical description of datasets."""
```
sbrugman, line 1: summary.py might be a better name for the file. Items in the description set are often statistical, but they do not have to be.

src/pandas_profiling/visualisation/plot.py
```diff
@@ -3,18 +3,17 @@
 from typing import Union, Optional
 
 import matplotlib
+import matplotlib.pyplot as plt
```
sbrugman, line 6: I see the logic in the other import changes, but why is `matplotlib.colors` not grouped with the other matplotlib imports here?

In the future, it is better to put these code style changes (organizing the imports) into a separate PR (for easier review).

src/pandas_profiling/config.py
```diff
@@ -66,5 +66,22 @@ def __setitem__(self, key, value):
...
+        if not isinstance(other, Config):
+            raise ValueError("Can only update config from a config object")
+        self.config = other.config
+
+    def clear(self):
+        self.config = confuse.Configuration("PandasProfiling", __name__, read=False)
+        self.set_file(str(get_config_default()))
+
+    @property
+    def is_default(self):
```
sbrugman, line 79: Seems to be untested

src/pandas_profiling/model/correlations.py
```diff
@@ -4,13 +4,12 @@
 from contextlib import suppress
 from functools import partial
 from typing import Callable, Dict, List, Optional
-
-import pandas as pd
+import phik
```
sbrugman, line 7: This import should remain at it's orginal location. It allows users to slim down the dependencies when this coefficient is not used.

sbrugman:
I've reviewed the code, you can find the comments above.

> BTW, The "dask branch" only uses dask to schedule the computation steps in the description method. I have ran some tests, so I considered it as an optional experimental feature.

Regarding to dask, have you noticed any performance gain? Preferably, add a benchmark. Furthermore, the dask branch is missing diagnostic information (such as a progress bar). This would greatly benefit our users. More information can be found here: https://docs.dask.org/en/latest/diagnostics-distributed.html. Finally, there should information added to the documentation/Readme to explain users when and why to use dask.

src/pandas_profiling/__init__.py
loopyme: If users instantiates two `ProfileReport` object in one process, they will share the config. This warning warns when user did not specify config, but `ProfileReport` implicitly used previous non-default config.

src/pandas_profiling/__init__.py
loopyme: I have considered about other methods of serialization and deserialization, the central paradox is that report and description_set contains many types of object.
 - Built-in types: `bool`, `int`, `str`, `float`, `None`: Can be handled by json
 - Custom types: `Message`,`Renderable`,`Sequence`: Can be handled by mixin class you have mentioned.
 - Enum: `MessageType`,`Variable`: Need some function and wrapper to handle it
 - Custom types in dependent package : `np.int64`,`np.float64`,`np.bool_`,`np.ndarray`,`pd.Series`,`datetime`: Need wrapper class to handle correspondingly.

Instead of pickle, I think the alternative is to design a special data structure instead of a dictionary to store and handle the data. I chose pickle because the usage of report and  description_set are so broad that changing the core data structure might cause instability.

src/pandas_profiling/model/describe.py
```diff
@@ -571,87 +63,80 @@ def describe(df: pd.DataFrame) -> dict:
...
+
+        pbar.set_postfix_str("Get variable types")
+        variables = {
+            column: description["type"]
+            for column, description in series_description.items()
+        }
+        pbar.update()
 
-            # TODO: use `Pool` for Linux-based systems
-            with multiprocessing.pool.ThreadPool(pool_size) as executor:
```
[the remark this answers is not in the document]
loopyme: I've moved the multithreading stuffs to `.model.summary.get_series_descriptions` to make the description process clearer.

src/pandas_profiling/model/describe.py
loopyme: I've noticed many use of the statement 'Describe', like ''description_set", "describe_df", so I think maybe 'describe' can better continue the previous naming style.

src/pandas_profiling/visualisation/plot.py
loopyme: In fact, I don't know. It seems that all import statements are sorted alphabetically, and this statement start with `from`. My IDE did this job automatically and I committed it because I think it looks good.

Thanks for your advice, I will put code style changes into a separate PR in the future.

loopyme:
Thanks for review!

Over the last few days, I've been refactoring report function, tweaking some low-level functions and testing the dask-scheduling performance gain on several benchmarks. According to the current results, dask-scheduling tends to achieve more performance gains when there are more variables in the data set, which may be due to the fact that dask-scheduling makes the most part of describe function parallel.

However, further parallel optimization seems impossible, because the plot function is not thread-safe and highly unstable when parallel computing. I think adjusting config will be another direction of performance improvement.

I'll add diagnostic information to it after finishing the present work.

src/pandas_profiling/__init__.py
sbrugman: For now pickle looks like the most stable option indeed.

PR #436, opened 2020-04-17
Added feature name to description bar

zkid18:
Adding feature name in progress bar panel could be a nice way to swiftly debug problems like that https://github.com/pandas-profiling/pandas-profiling/issues/434


zkid18:
Haven't noticed that @loopyme have made the same PR a few minutes before https://github.com/pandas-profiling/pandas-profiling/pull/435

PR #465, opened 2020-05-09
Each ProfileReport has its own configuration

loopyme:


loopyme:
As we discussed before, module-level config are used not only in model.summary, model.describe but also in controller, model.base, model.correlations, model.message,variables rendering and many more  file(23 files, actually) with more than 150 value read. Passing config reference every where will confuse the structure.

Register-config way implement in this PR compromises the old fashion with the current design, which enables each ProfileReport to have its own configuration (user perspective) whether they are run in parallel or serial

Issue #471, opened 2020-05-13
density plots option for interactions


Iuiu1234:
When plotting interactions between a numerical (x) vs categorical variable (y).
There is the option to plot it as density plot of x grouped by y? Using seaborn kdeplot or similar?

(![Figure_1](https://user-images.githubusercontent.com/30587996/81862418-fe5f8900-9569-11ea-915a-4791f06c3c76.png))

sbrugman:
On the roadmap, but PR's welcome.

sbrugman:
We've found that for the majority of the use cases, estimating the kde is computationally too expensive. Favouring #344 for visualising pairwise relationships atm.

PR #472, opened 2020-05-13
density-plot

Iuiu1234:
Option to solve issue #471.


sbrugman:
Thanks for your PR. I've tried to include this, however, KDE is seems too expensive for many use cases, hence we'll include a computationally less expensive method for pairwise relations between categorical variables: heatmaps.

I'm closing this PR. You're very much welcome to join us in discussing and developing the new approach!

PR #501, opened 2020-06-16
Add matplotlib unit registry and rcParams context manager

### LINKED ISSUE #494: pandas_profiling modifies heavily matplotlib's rcParams
Thanks for this nice library.

Importing pandas_profiling modifies heavily matplotlib's rcParams. As a result it may render other plots unreadable (see for example
https://github.com/scikit-learn/scikit-learn/issues/17542 )

It would be much nicer if pandas_profiling set directly the desired tweaks on the visualizations that it creates, rather than modifying the global settings.

Another option would be to use the rc_context context manager: https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.rc_context.html It might be easier to code, but the drawback is that it would not be thread safe.

Thanks!

Jooong:
### Summary
* Fixes #494 
* Added context manager to avoid pandas-profiling permanently change the global context of matplotlib. The context manager modifies and rolls back  `matplotlib.rcParams` and `matplotlib.inits.Registry`. Every plotting function is decorated with the context manager. 
* Refactored mplstyle file into a python dictionary

It would be grateful if someone spend time to review this patch and let me know if there's any possible issues with this patch or other suggestions.


Jooong:
Noticed using context manager is not thread safe. Looking for other workaround.

Jooong:
I think current patch is okay. Found that thread pool is used to run [describe_1d](https://github.com/pandas-profiling/pandas-profiling/blob/be8baee64cebb49aea187fa05efe8ba84e7823c8/src/pandas_profiling/model/summary.py#L49), calculating statistics from pandas series. The visualization functions are currently run in a single thread. If you still think it can be a problem in the future, please let me know.

sbrugman:
Thanks for this workaround. Not experiencing side effects on the matplotlib rcparams will be much appreciated by users.

I've got two points we should consider before merging:
* In this implementation the params are set and restored at every plotting call. I'd like to know how much overhead this adds. If it's not neglectible, we might add the context manager elsewhere. For instance, when the report is rendered.
* Seaborn is currently not a dependency in any of the releases. If we can, i'd like to avoid making it mandatory. The fact that it is present in the `develop` branch is somewhat misleading, as it's not expected to be in the next release.

Jooong:
Hi, thanks for your review. I've done a naive profiling using [examples/bank_marketing_data/banking_data.py](https://github.com/Jooong/pandas-profiling/blob/4d841c71f9cda3f440df0930234bf924efc77b79/examples/bank_marketing_data/banking_data.py). Based on this [commit](https://github.com/Jooong/pandas-profiling/commit/0e9093835550bf9fc91598048a061d961f716855), total overhead of `manage_matplotlib_context` was **0.062** second, while executing `to_file` method of `ProfileReport` took total **14.2507** second. Average overhead of calling this function was **0.00081** second. I think it's acceptable overhead. If you think this isn't acceptable or this experiment is insufficient, please let me know.

PR #514, opened 2020-07-09
Tiny enhancement in the Support section of the documentation

smaranjitghose:


sbrugman:
Thanks for thinking along.

Now that this page has our attention, we can refine it a bit further.
- It seems to make more sense to use inline code markup for each of the commands individually
- We also would like to be consistent for the commands mentioned earlier

Proposed:
>             - package manager (e.g. ``pip --version`` or ``conda info``)
>             - packages (e.g. ``pip freeze > packages.txt`` or ``conda list``)
>        - a sample and/or description of the dataset ``df.head()`` and ``df.info()``

sbrugman:
Thank you for processing the feedback @smaranjitghose.

Minor changes requested before we can merge:
- ``pip freeze > packages.txt`` is a single command and should be denoted as such (the last part instructs writing the output of pip freeze to the `packages.txt` file). 
- For ``df.head(), df.info()``, we have exactly the opposite case. The command should be separated.

Example:
>         - a sample and/or description of the dataset ``df.head()`` and ``df.info()``

PR #522, opened 2020-07-18
Add column definition function as a separate tab

mereldawu:
## What was changed:

Added "Column definitions" in the generated report as a tab after Duplicated rows. This tab shows the column definitions of the dataset as a table using the striped table format.

## Why was it changed:

Our team has multiple data scientists, and we use pandas-profiling to generate exploratory analysis reports and share them among the team members and stakeholders. We want to distribute one file, instead of distributing both the pandas-profiling report and a separate column definitions file.

## How was it changed:

Added column definitions into `describe.py`. 

It is set as True in the `config_default.yaml` and False in the `config_minimal.yaml`.

The ProfileReport constructor takes a new parameter (`definition_file`). The path to the column definition file is passed in as the argument. Currently, only the JSON file format is implemented, and It defaults to None. 

The example is implemented in `examples/census/census.py` and the column definition is in `examples/census/census_column_definition.json`. 

```
profile = ProfileReport(
    df,
    title="Census Dataset",
    explorative=True,
    definition_file=f'{filepath}/census_column_definition.json'
)
```

The required dataset_column_definition.json format is:
```
{
    "column name 1": "column 1 definition",
    "column name 2": "column 2 definition"
}
```


sbrugman:
@mereldawu thank you for this pull request.

`pandas-profiling` already has the option to add descriptions to individual columns (see [this example](https://pandas-profiling.github.io/pandas-profiling/docs/master/rtd/pages/advanced_usage.html#id2)). The difference between the current implementation and your work is that the descriptions are (1) loaded from a file and (2) displayed in their own section.

Loading the descriptions from a file can be easily achieved in the current implementation like this:
```python
import json
import pandas as pd
import pandas_profiling


with open('census_column_definition.json', 'r') as f:
    definitions= json.load(f)

report = df.profile_report(variable=dict(descriptions=definitions))
report.to_file('report.html')
```
 
The second difference would be useful to incorporate, since his provides a better overview of the descriptions. Would you be interested in changing the PR in such a way that the descriptions are provides as a tab under "Overview"?

See https://github.com/pandas-profiling/pandas-profiling/blob/develop/src/pandas_profiling/report/structure/overview.py

mereldawu:
Hey @sbrugman, 

I didn't know that there was already a variable description option in place!

In any case, I moved the implementation to the Overview tab and updated the example as you suggested.
Apologies for the multiple commits, I didn't see your edit so had to add the example separately.

sbrugman:
@mereldawu Thanks for your contribution. I've changed a few things:

- Rather than using two sources for the descriptions, we now use the `variables.descriptions` parameter for both and have an additional parameter to disable displaying the descriptions for each variable.
- Restructured part of the overview to be consistent with the changes.

Could you review?

PR #524, opened 2020-07-19
Update Streamlit integration

Ghasel:
As suggested in https://github.com/Ghasel/streamlit-pandas-profiling/issues/2, here is an doc update on Pandas Profiling integration in Streamlit.


sbrugman:
LGTM, thanks for the quick PR.

Issue #537, opened 2020-08-06
ValueError: shape mismatch: value array of shape (136,) could not be broadcast to indexing result of shape (135,)


SurajitTest:
Hi,
     I have a dataset with 42 numerical features and 143 categorical features. I am on pandas 1.1.0 and pandas-profiling 2.8.0. I did try to use pandas-profiling with the given below commands:

profile=pandas_profiling.ProfileReport(X_train_After_FS_1)
profile.to_file("my_filename.html")

**Error Details :**
HBox(children=(FloatProgress(value=0.0, description='Summarize dataset', max=200.0, style=ProgressStyle(descri…

Traceback (most recent call last):

  File "<ipython-input-13-d112e0e4de59>", line 2, in <module>
    profile.to_file("C:\\Users\\ctystq\\Surajit_Personal\\AMPLifeModels\\Transformation_Project\\LS_Lapse_Propensity\\2.Category-2_RS _Ord_Models\\M1-ANVRSY_MTHS_NON_TR\\EDA_Output\\Plots\\Dataset_Profile_Raw_Data_From_Hive.html")

  File "D:\Installed-sw\Anaconda3\envs\root_clone-Hive\lib\site-packages\pandas_profiling\profile_report.py", line 245, in to_file
    data = self.to_html()

  File "D:\Installed-sw\Anaconda3\envs\root_clone-Hive\lib\site-packages\pandas_profiling\profile_report.py", line 348, in to_html
    return self.html

  File "D:\Installed-sw\Anaconda3\envs\root_clone-Hive\lib\site-packages\pandas_profiling\profile_report.py", line 168, in html
    self._html = self._render_html()

  File "D:\Installed-sw\Anaconda3\envs\root_clone-Hive\lib\site-packages\pandas_profiling\profile_report.py", line 275, in _render_html
    report = self.report

  File "D:\Installed-sw\Anaconda3\envs\root_clone-Hive\lib\site-packages\pandas_profiling\profile_report.py", line 162, in report
    self._report = get_report_structure(self.description_set)

  File "D:\Installed-sw\Anaconda3\envs\root_clone-Hive\lib\site-packages\pandas_profiling\profile_report.py", line 143, in description_set
    self._description_set = describe_df(self.title, self.df)

  File "D:\Installed-sw\Anaconda3\envs\root_clone-Hive\lib\site-packages\pandas_profiling\model\describe.py", line 63, in describe
    series_description = get_series_descriptions(df, pbar)

[...]

PR #548, opened 2020-08-18
fix ValueError race condition when running multiprocessing with describe1d

chanedwin:
References issue #537 

Problem :
ValueError is raised when running ProfileReport on large datasets and with multiprocessing on (pool_size >1). This is likely due to the `series.fillna(np.nan, inplace=True)`  in summary.py seems to be performing multiple in-place mutations to the underlying DataFrame object through the passed series reference, resulting in some kind of race condition where two of the processes try to write to the DataFrame at the same time and the ValueError then occurs. This is also why changing the pool_size to 1 fixes the issue, and why the error doesn't always occur - you probably need enough data and threads to hit the race condition. 

Solution : 
Replace `series.fillna(np.nan, inplace=True)` with `series = series.fillna(np.nan)` , negating any side effects from mutation.

### DIFF
--- a/src/pandas_profiling/model/summary.py
+++ b/src/pandas_profiling/model/summary.py
@@ -427,7 +427,7 @@ def describe_1d(series: pd.Series) -> dict:
         return stats
 
     # Make sure pd.NA is not in the series
-    series.fillna(np.nan, inplace=True)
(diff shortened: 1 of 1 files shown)


chanedwin:
closing until tests are fixed

PR #590, opened 2020-09-27
Restructuring categorical overview

sbrugman:
- Condense frequency tables
- Include word frequency statistics
- Restructure the overview


neomatrix369:
@sbrugman I couldn't help notice there are a few text statistics included in this bit of functionality.
Do you see any blending points between pandas profiling and [NLP Profiler](https://github.com/neomatrix369/nlp_profiler/)? Happy to have a discussion on this if you like, after you have given it a thought.

sbrugman:
@neomatrix369 That's right. The character statistics are powerful when analyzing text sequences, as the unicode categories and other properties reveal many interesting aspects of the data (case, punctuation, spacing, special characters as emoji's).

The statistics in this package are meant to be general purpose. Features such as spelling checks are highly specialized (the value of the analysis is dependent on the quality of the language model and is language-specific). Those kind of analysis require broader control by the user, while unicode for instance should be deterministic.

pandas-profiling v3 aims to include bring-your-own (BYO) plots and statistics. This allows you to perform use-case specific analyses.

neomatrix369:
> @neomatrix369 That's right. The character statistics are powerful when analyzing text sequences, as the unicode categories and other properties reveal many interesting aspects of the data (case, punctuation, spacing, special characters as emoji's).
> 
> The statistics in this package are meant to be general purpose. Features such as spelling checks are highly specialized (the value of the analysis is dependent on the quality of the language model and is language-specific). Those kind of analysis require broader control by the user, while unicode for instance should be deterministic.
> 
> pandas-profiling v3 aims to include bring-your-own (BYO) plots and statistics. This allows you to perform use-case specific analyses.

What if there are text fields (they will be `object` type and not `category` type) does it also perform this analysis on such fields?

In my view, an analysis/profiling tool should provide both high level and low-level (granular) details and leave the option to the user to choose from. 

NLP Profiler has both of these, by details you get it all and they can be toggled with a flag.

PR #617, opened 2020-11-11
add bar2 to missing diagrams (missing bar count)

studyproject-mlds:
add missing bar count to missing diagrams

image


sbrugman:
@studyproject-mlds I'm missing what adding a new diagram with the counts adds. Could you please elaborate?

PR #619, opened 2020-11-14
BUG: Make sure nans are converted to empty strings (closes #613)

### LINKED ISSUE #613: NaN value is treated as a string in length frequency analysis
It seems to me that a missing value (`NaN`) is treated as a string of 3 characters in the _Histogram of lengths of the category_ feature. I believe this behaviour is misleading, and thus would call it a bug.

**And example to be reproduced**

```
import numpy as np
import pandas as pd
df = pd.DataFrame([{'col': 'ABCD'}, {'col': 'DEFG'}, {'col': np.nan}])
ProfileReport(df)
```

Yields the following length statistics of the column `col`, which do not seem to be correct. 

Mean length = 3.666666667
Min length = 3

Or, graphically:

image

IMHO the empty value should not be taken into account in this kind of analysis.

**Version information**

* _Python version_: 3.8.3
* _Environment_: Local Jupyter Notebook
* pandas_profiling 2.9.0
* pandas 1.0.5
* numpy 1.18.5

SunnyxBd:
NaN values are being converted to string "nan"

I fixed this issue by making sure Nans were replaced by empty strings first

### DIFF
--- a/src/pandas_profiling/model/summary.py
+++ b/src/pandas_profiling/model/summary.py
@@ -286,7 +286,8 @@ def describe_1d(series: pd.Series) -> dict:
             A dict containing calculated series description values.
         """
         # Make sure we deal with strings (Issue #100)
-        series = series.astype(str)
+        # Make sure nans are converted to empty strings (Issue #613)
+        series = series.fillna('').astype('str')
 
         # Only run if at least 1 non-missing value
         value_counts = series_description["value_counts_without_nan"]
--- a/tests/issues/test_issue613.py
+++ b/tests/issues/test_issue613.py
@@ -1,0 +1,17 @@
+"""
+Test for issue 613:
+https://github.com/pandas-profiling/pandas-profiling/issues/613
+"""
+import numpy as np
+import pandas as pd
+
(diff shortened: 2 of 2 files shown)


sbrugman:
@SunnyxBd Great first attempt at resolving this, thank you. 

From the tests we can see that the current solution does not work with pandas' categorical types ("ValueError: fill value must be in categories"). A way to work around this is to first take the the missing mask (e.g. `series.isna()`), then convert to string and then replace the missing values with the empty string.

sbrugman (changes requested):
Otherwise looks good. 

You might want to lint the code (`make lint`)

sbrugman:
@SunnyxBd Would you still be interested to contribute this fix? If you need any pointers, feel free to ask.

sbrugman:
A fix for this issue is included in the new release. Feel free to take on one of the [getting started ☝](https://github.com/pandas-profiling/pandas-profiling/issues?q=is%3Aissue+is%3Aopen+label%3A%22getting+started+%E2%98%9D%22) issues if you're interested in contributing later.

PR #649, opened 2020-12-24
Handle on the fly decompression of data files supported by pandas (#652)

scholarsmate:
Pandas can handle on the fly decompression from the following extensions: ‘.bz2’, ‘.gz’, ‘.zip’, or ‘.xz’ (otherwise no decompression). Now the profiler can as well. (#652)

Using LZMA (.xz) typically yields a compression ratio of 25:1 for my very large .csv datasets (many GBs).  Being able to operate on them compressed with on the fly support in pandas has been extremely beneficial.  This patch extends these same benefits to the profiler.  I think others will benefit as well.


sbrugman:
@scholarsmate Thanks for contributing this. Could you please add a unit test?

PR #666, opened 2021-01-10
integration with great expectations

sbrugman:
Bootstrap the great-expectations setup for a dataset based on the semantic data types and metrics computed by pandas-profiling.

(includes minor bugfixes for json representation and slug)

### DIFF
--- a/docsrc/source/pages/great_expectations_integration.rst
+++ b/docsrc/source/pages/great_expectations_integration.rst
@@ -88,10 +88,15 @@ The ``to_expectation_suite`` method returns a default set of Expectations if Pan
 * ``expect_column_values_to_be_in_type_list``
 * ``expect_column_values_to_be_increasing``
 * ``expect_column_values_to_be_decreasing``
+* ``expect_column_values_to_be_between``
 
 **Categorical columns**
 
 * ``expect_column_values_to_be_in_set``
+
+**Datetime columns**
+
+* ``expect_column_values_to_be_between``
 
 **Filename columns**
 
--- a/docsrc/source/pages/great_expectations_integration.rst
+++ b/docsrc/source/pages/great_expectations_integration.rst
@@ -68,15 +68,14 @@ You can also configure each feature individually in the function call:
         handler=handler
     )
 
-TODO: explain the handler
-
 See `the Great Expectations Examples <https://pandas-profiling.github.io/pandas-profiling/examples/master/features/great_expectations_example.html>`_ for complete examples.
 
 
 Included Expectation types
 --------------------------
 
-The ``to_expectation_suite`` method returns a default set of Expectations if Pandas Profiling determines that the assertion holds true for the profiled dataset. The Expectation types depend on the datatype of a column:
+The ``to_expectation_suite`` method returns a default set of Expectations if Pandas Profiling determines that the assertion holds true for the profiled dataset.
+The Expectation types depend on the datatype of a column:
 
 **All columns**
 
@@ -101,3 +100,51 @@ The ``to_expectation_suite`` method returns a default set of Expectations if Pan
 **Filename columns**
 
 * ``expect_file_to_exist``
+
+
+The default logic is straight forward and can be found here in `expectation_algorithms.py <https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/model/expectation_algorithms.py>`_.
+
+Rolling your own Expectation Generation Logic
+---------------------------------------------
+
+If you would like to profile datasets at scale, your use case might require to change the default expectations logic.
+The ``to_expectation_suite`` takes the ``handler`` parameter, that allows you to take full control of the generation process.
+Generating expectations takes place in two steps:
+
+- mapping the detected type of each column to a generator function (that receives the columns' summary statistics);
+- generating expectations based on the summary (e.g. ``expect_column_values_to_not_be_null`` if ``summary["n_missing"] == 0``)
+
+Adding an expectation to columns with constant length can be achieved for instance using this code:
+
+.. code-block:: python
+
+    def fixed_length(name, summary, batch, *args):
+        """Add a length expectation to columns with constant length values"""
+        if summary["min_length"] == summary["max_length"]:
+            batch.expect_column_value_lengths_to_equal(summary["min_length"])
+        return name, summary, batch
+
+
+    class MyExpectationHandler(Handler):
+        def __init__(self, typeset, *args, **kwargs):
+            mapping = {
+                Unsupported: [expectation_algorithms.generic_expectations],
+                Categorical: [expectation_algorithms.categorical_expectations, fixed_length],
+                Boolean: [expectation_algorithms.categorical_expectations],
+                Numeric: [expectation_algorithms.numeric_expectations],
+                URL: [expectation_algorithms.url_expectations],
+                File: [expectation_algorithms.file_expectations],
+                Path: [expectation_algorithms.path_expectations],
+                DateTime: [expectation_algorithms.datetime_expectations],
+                Image: [expectation_algorithms.image_expectations],
+            }
+            super().__init__(mapping, typeset, *args, **kwargs)
+
+    # (initiate report)
+
+    suite = report.to_expectation_suite(
+        handler=MyExpectationHandler(report.typeset)
+    )
+
+You can automate even more by extending the typeset (by default the ``ProfilingTypeSet``) with semantic data types specific to your company or use case (for instance disease classification in healthcare or currency and IBAN in finance).
+For that, you can find details in the `visions <https://github.com/dylan-profiler/visions>`_ documentation.
--- a/docsrc/source/pages/great_expectations_integration.rst
+++ b/docsrc/source/pages/great_expectations_integration.rst
@@ -1,0 +1,98 @@
+====================================
+Integration with Great Expectations
+====================================
+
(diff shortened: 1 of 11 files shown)


src/pandas_profiling/expectations_report.py
```diff
@@ -0,0 +1,91 @@
...
+        batch = ge.dataset.PandasDataset(self.df, expectation_suite=suite)
+
+        # Obtain the profiling summary
+        summary = self.get_description()
+
+        # Dispatch to expectations per semantic variable type
+        for name, variable_summary in summary["variables"].items():
+            handler.handle(variable_summary["type"], name, variable_summary, batch)
+
+        return batch
```
talagluck, line 67:
```suggestion
        return batch.get_expectation_suite()

```

src/pandas_profiling/expectations_report.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def to_expectations(self, suite_name=None, handler=None, build_data_docs=True):
```
talagluck, line 69: Maybe we can instead call this something like `to_expectations_with_validation_operator` to differentiate a bit more clearly

src/pandas_profiling/model/expectation_algorithms.py
```diff
@@ -0,0 +1,70 @@
+def generic_expectations(name, summary, batch, *args):
+    batch.expect_column_to_exist(name)
+
+    if summary["n_missing"] == 0:
+        batch.expect_column_values_to_not_be_null(name)
```
talagluck, line 5: Just a heads-up that it is possible to use the `mostly` parameter here, so for instance, if the % missing was less than 10%, you could use `batch.expect_column_values_to_not_be_null(name, mostly=0.9)`. I would say it's your call as to whether to use this, and what the threshold should be if you do.

src/pandas_profiling/model/expectation_algorithms.py
```diff
@@ -0,0 +1,70 @@
...
+    if summary["n_missing"] == 0:
+        batch.expect_column_values_to_not_be_null(name)
+
+    if summary["p_unique"] == 1.0:
+        batch.expect_column_values_to_be_unique(name)
+
+    return name, summary, batch
+
+
+def numeric_expectations(name, summary, batch, *args):
```
talagluck, line 13: I think it's worth adding a few more numeric expectations here, like `expect_column_values_to_be_between` or `expect_column_min_to_be_between` and `expect_column_max_to_be_between`. You could add in a tolerance (e.g. adding 10% to a column max and subtracting 10% from a column min), but I think having the expectation be very exact is also OK.

src/pandas_profiling/model/expectation_algorithms.py
```diff
@@ -0,0 +1,70 @@
...
+
+    if summary["monotonic_decrease"]:
+        batch.expect_column_values_to_be_decreasing(
+            name, strictly=summary["monotonic_decrease_strict"]
+        )
+
+    return name, summary, batch
+
+
+def categorical_expectations(name, summary, batch, *args):
```
talagluck, line 42: I think it would be worth adding an `expect_column_values_to_be_in_set` for categorical columns. There is subjectivity with regard to what is considered categorical, whether it is an absolute number or a percentage of column value count (and we are working through this with our own internal profilers as well). I think deciding on a threshold and going with that will be fine.

src/pandas_profiling/model/expectation_algorithms.py
```diff
@@ -0,0 +1,70 @@
...
+    return name, summary, batch
+
+
+def boolean_expectations(name, summary, batch, *args):
```
talagluck, line 46: This should be able to take an `expect_column_values_to_be_in_set` as well.

src/pandas_profiling/model/expectation_algorithms.py
```diff
@@ -0,0 +1,70 @@
...
+    return name, summary, batch
+
+
+def path_expectations(name, summary, batch, *args):
+    return name, summary, batch
+
+
+def datetime_expectations(name, summary, batch, *args):
```
talagluck, line 54: This might be a good candidate for `expect_column_values_to_be_between` or at least `expect_column_min_to_be_between`. For `expect_column_values_to_be_between`, you can leave out the `max_value` to have a lower bound and no upper bound.

talagluck (commented):
Overall, this looks great! Just a few suggestions, mostly around expectations to add.

With regard to adding expectations, the thing to keep in mind is that the expectations should be sensible and transparent, but they don't need to be perfect - they will likely need some adjustment after the initial suite creation. This is more about creating an easy starting point for users to build an expectation suite, and less about building a perfect final suite for users.

src/pandas_profiling/expectations_report.py
```diff
@@ -0,0 +1,91 @@
...
+            Path: [expectation_algorithms.path_expectations],
+            Boolean: [expectation_algorithms.boolean_expectations],
+            DateTime: [expectation_algorithms.datetime_expectations],
+            Image: [expectation_algorithms.image_expectations],
+        }
+        super().__init__(mapping, typeset, *args, **kwargs)
+
+
+class ExpectationsReport:
+    def to_expectations_suite(self, suite_name=None, handler=None):
```
talagluck, line 36: A nitpick: Could we change this to `to_expectation_suite`?

src/pandas_profiling/expectations_report.py
```diff
@@ -0,0 +1,91 @@
...
+            import great_expectations as ge
+        except ImportError:
+            raise ImportError(
+                "Please install great expectations before using the expectation functionality"
+            )
+
+        context = ge.data_context.DataContext()
+
+        batch = self.to_expectations_suite(suite_name, handler)
+
```
talagluck, line 80:
```suggestion

# This validation operator works with great_expectations <0.12 API, and will need migrating with new style batches and validator
```

src/pandas_profiling/expectations_report.py
```diff
@@ -0,0 +1,91 @@
...
+        # Use report title if suite is empty
+        if suite_name is None:
+            suite_name = slugify(self.title)
+
+        # Use the default handler if none
+        if handler is None:
+            handler = ExpectationHandler(ProfilingTypeSet())
+
+        # Obtain the ge context and create the expectation suite
+        context = ge.data_context.DataContext()
```
talagluck, line 53: One suggestion from Sam - we've recently been improving our in-memory data contexts, and we're starting to get users who are building data contexts in that way. If we allow users to pass a data context into the function (and only instantiate a data context if nothing is passed in), then it would also support those users.

src/pandas_profiling/expectations_report.py
talagluck, line 69: Could we add a return to this function as well? Perhaps just a list of the expectations created.

talagluck (commented):
And a few more comments!

docsrc/source/pages/great_expectations_integration.rst
```diff
@@ -0,0 +1,150 @@
...
+
+For more information about Great Expectations, check out the `Great Expectations documentation <https://docs.greatexpectations.io/en/latest/>`_ and join the `Great Expectations Slack channel <https://www.greatexpectations.io/slack>` for help.
+
+
+Creating Expectation Suites with Pandas Profiling
+--------------------------------------------------
+
+An *Expectation Suite* is simply a set of Expectations. You can create Expectation Suites by writing out individual statements, such as the one above, or by automatically generating them based on profiler results.
+
+Pandas Profiling provides a simple ``get_expectation_suite()`` method that returns a Great Expectations ``ExpectationSuite`` object which contains a set of Expectations.
```
talagluck, line 27:
```suggestion
Pandas Profiling provides a simple ``to_expectation_suite()`` method that returns a Great Expectations ``ExpectationSuite`` object which contains a set of Expectations.
```

docsrc/source/pages/great_expectations_integration.rst
```diff
@@ -0,0 +1,150 @@
...
+
+**Pre-requisites**: In order to run the ``get_expectation_suite()`` method, you will need to install Great Expectations:
```
talagluck, line 29:
```suggestion
**Pre-requisites**: In order to run the ``to_expectation_suite()`` method, you will need to install Great Expectations:
```

docsrc/source/pages/great_expectations_integration.rst
```diff
@@ -0,0 +1,150 @@
...
+`pip install great_expectations`
+
+If you would like to use the additional features such as saving the Suite and building Data Docs, you will also need to configure a Great Expectations Data Context by running ``great_expectations init``.
```
talagluck, line 32:
```suggestion
If you would like to use the additional features such as saving the Suite and building Data Docs, you will also need to configure a Great Expectations Data Context by running ``great_expectations init`` while in your project directory.
```

docsrc/source/pages/great_expectations_integration.rst
```diff
@@ -0,0 +1,150 @@
...
+
+* ``expect_file_to_exist``
+
+
+The default logic is straight forward and can be found here in `expectation_algorithms.py <https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/model/expectation_algorithms.py>`_.
+
+Rolling your own Expectation Generation Logic
+---------------------------------------------
+
+If you would like to profile datasets at scale, your use case might require to change the default expectations logic.
```
talagluck, line 110:
```suggestion
If you would like to profile datasets at scale, your use case might require changing the default expectations logic.
```

docsrc/source/pages/great_expectations_integration.rst
```diff
@@ -0,0 +1,150 @@
...
+The ``to_expectation_suite`` takes the ``handler`` parameter, that allows you to take full control of the generation process.
```
talagluck, line 111:
```suggestion
The ``to_expectation_suite`` takes the ``handler`` parameter, which allows you to take full control of the generation process.
```

talagluck (approved):
Looks good! Just a few minor wording suggestions/corrections.

Issue #655, opened 2021-01-04
The "conda install defaults to 1.4.1" saga continues ...


CodeCox:
Not a bug but related to #448 and others as per the [FAQ](https://pandas-profiling.github.io/pandas-profiling/docs/master/rtd/pages/support.html#frequent-issues)

The ffg advice did not work for me!

    conda install -c conda-forge pandas-profiling=2.9.0

The command fails with 

    UnsatisfiableError: The following specifications were found to be incompatible with each other:

but conda does not actually list the incompatible dependancies. I found this rather unhelful since it gave me no leads to follow. _(For details, see  [issue #10443](https://github.com/conda/conda/issues/10443#issue-777683222) I raised in the conda repo about this lack of information)_
 
___

Anyway, eventually I was able to get `v2.9.0` by rolling back python `v3.9.1` to` v3.8.6`. This info in itself is not useful since obviously individual env's will differ. However I would like to draw your attention to the **`mamba` package manager** which helped me to _troubleshoot_ and _pinpoint_ the incompatible dependancy. I'll list a summarised version of my steps to illustrate my process since it may be a means of getting a more robust fix to this common issue since it does not resort to the magic of creating a fresh new env, etc:

- I downloaded mamba
 ```
conda install mamba
```
- This is the version I'm using
```
mamba list mamba 
mamba                     0.7.6            py39ha0edc35_0    conda-forge
```
- I try a dry-run install of the latest pandas-profiling
```
mamba install pandas-profiling=2.9.0 -d
Encountered problems while solving.
Problem: package pandas-profiling-2.9.0-pyh9f0ad1d_0 requires phik >=0.9.10, but none of the providers can be installed
```
- I try a dry-run install of the package listed in previous error 
```
mamba install phik -d
Encountered problems while solving.
Problem: package phik-0.10.0-py_0 requires numba >=0.38.1, but none of the providers can be installed
```
- I try a dry-run install of the package listed in previous error 
```
mamba install numba -d
Encountered problems while solving.
Problem: nothing provides numpy >=1.10,<1.11.0a0 needed by numba-0.36.1-np110py27h67d0f83_0
```
- But I know I have numpy! I notice the numpy version requirement above looks very old so I check my version
```mamba list numpy
numpy                     1.19.4           py39h6635163_1    conda-forge
```
- And I also see the mismatch with numba requirement ie 0.36 vs >=0.38 so I check for the latest version
```
mamba search numba
[...]

PR #667, opened 2021-01-10
doc: added #655 to support faq

CodeCox:
Added `mamba` troubleshooting tip from #655  to the support faq doc.


sbrugman:
Thanks!

PR #670, opened 2021-01-14
WIP: develop spark profiling

chanedwin:


chanedwin:
> This pull request **introduces 9 alerts** when merging [9a62ad4](https://github.com/pandas-profiling/pandas-profiling/commit/9a62ad4311cf96e874d6f3a34d68ab34dc7db921) into [f8333d7](https://github.com/pandas-profiling/pandas-profiling/commit/f8333d71e3887e5f7b7b6132bbecd2eadffcc7d4) - [view on LGTM.com](https://lgtm.com/projects/g/pandas-profiling/pandas-profiling/rev/pr-2f60600705321e3d6e91738ddb630687596baa44)
> 
> **new alerts:**
> 
> * 4 for Unused import
> * 3 for Variable defined multiple times
> * 1 for Except block handles 'BaseException'
> * 1 for Module is imported with 'import' and 'import from'

Fixed in commit 0921156

src/pandas_profiling/config_default.yaml
```diff
@@ -77,25 +77,39 @@ missing_diagrams:
 correlations:
     pearson:
       calculate: True
+      spark_calculate: True
```
sbrugman, line 80: I wouldn't introduce calculate parameters for each engine (pandas/spark). Rather have a single flag for engine. For spark some correlations won't be implemented, for which we can ignore the configuration alltogether.

src/pandas_profiling/model/describe.py
```diff
@@ -44,16 +50,39 @@ def describe(
...
+        else:
+            df.persist_bool = False
+
     disable_progress_bar = not config["progress_bar"].get(bool)
 
     date_start = datetime.utcnow()
 
+    # if there are specific config for each correlation, then use those
+
+    if isinstance(df, SparkDataFrame):
```
sbrugman, line 81: Obsolte if we use the change proposed above

Makefile
```diff
@@ -12,18 +12,21 @@ docs:
...
-	pytest --cov=. --cov-append tests/issues/
-	pytest --cov=. --cov-append --nbval tests/notebooks/
+	pytest -m "not sparktest" --cov=. tests/unit/
+	pytest -m "not sparktest" --cov=. --cov-append tests/issues/
+	pytest -m "not sparktest" --cov=. --cov-append --nbval tests/notebooks/
 	pandas_profiling -h
 	make typing
 
+test-spark:
+	pytest -m sparktest tests/unit/
```
sbrugman, line 28: Shouldn't this have the --cov=. flag?

.github/workflows/ci_test.yml
```diff
@@ -63,3 +63,73 @@ jobs:
...
+#            pandas: ">1.1"
+#          - os: windows-2016
+#            python-version: 3.6
+#            pandas: ">1.1"
+
+
+
+    name: CI Tests Spark (py${{ matrix.python-version }}, ${{ matrix.os }}, pandas${{ matrix.pandas }}, spark${{ matrix.spark }})
+    env:
+      TRAVIS_BUILD_DIR: ${{ github.workspace }}
```
sbrugman, line 93: TRAVIS_BUILD_DIR => BUILD_DIR

Makefile
```diff
@@ -37,6 +40,12 @@ pypi_package:
 install:
 	pip install -e .[notebook]
 
+install-spark-ci:
+	sudo apt-get -y install openjdk-8-jdk
+	curl https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz \
+	--output ${TRAVIS_BUILD_DIR}/spark.tgz
+	tar -xvzf ${TRAVIS_BUILD_DIR}/spark.tgz && mv spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION} spark
```
sbrugman, line 47: The last "spark" isn't in the ${TRAVIS_BUILD_DIR}, right? Perhaps ${SPARK_HOME} would be better here (had some issues with this at another project).

pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[tool.pytest.ini_options]
+markers =  ["sparktest",]
```
sbrugman, line 2: `,` isn't required (toml supports arrays with a single element)

src/pandas_profiling/config_default.yaml
chanedwin: spark_calculate removed as config param

src/pandas_profiling/model/describe.py
chanedwin: removed as obsolete

Makefile
chanedwin: I kept all the unit tests separate (one set of spark, one set for pandas for all the matrix of parameters) and finally one full set of tests (pandas + spark) to measure coverage

chanedwin: see if this is ok? [tests](https://github.com/chanedwin/pandas-profiling/blob/develop-clone/.github/workflows/ci_test.yml) | [output](https://github.com/pandas-profiling/pandas-profiling/actions/runs/510768032)

.github/workflows/ci_test.yml
chanedwin: TRAVIS_BUILD_DIR removed as redundant

pyproject.toml
chanedwin: , removed

Makefile
chanedwin: replaced with SPARK_DIRECTORY variable for clearer naming

Makefile
sbrugman: In that approach one combination runs twice. As long as Github Action's runners are so generously provided this shouldn't be a problem, but there is room too optimize. Fine with me for now.

src/pandas_profiling/model/typeset.py
```diff
@@ -191,10 +192,50 @@ def get_relations(cls):
...
+
+class SparkNumeric(visions.VisionsBaseType):
+    @classmethod
+    def get_relations(cls):
+        return [IdentityRelation(cls, SparkUnsupported)]
+
+    @classmethod
+    def contains_op(cls, series: SparkSeries, state: dict) -> bool:
+        return (
+            str(series.type) == "DoubleType"
```
sbrugman, line 211: `str(series.type) in ["DoubleType", ...]` would be cleaner and faster

src/pandas_profiling/profile_report.py
```diff
@@ -92,8 +91,17 @@ def __init__(
         self._summarizer = None
 
         if df is not None:
-            # preprocess df
-            self.df = self.preprocess(df)
+            # get appropriate backend
+            df_wrapper = get_appropriate_wrapper(df)
```
sbrugman, line 95: get_execution_engine?

src/pandas_profiling/utils/common.py
```diff
@@ -88,3 +95,41 @@ def convert_timestamp_to_datetime(timestamp: int) -> datetime:
...
+
+        # if spark version == 2.3 or 2.4
+        if spark_version[0] == "2" and spark_version[1] in ["3", "4"]:
+
+            pyarrow_version = pyarrow.__version__.split(".")
+
+            # this checks if the env has an incompatible arrow version (not < 0.15)
+            if not (pyarrow_version[0] == "0" and int(pyarrow_version[1]) < 15):
+                # if os variable is not set, likely unhandled
+                if "ARROW_PRE_0_15_IPC_FORMAT" not in os.environ:
```
sbrugman, line 126: Can be simplified as:
```
if "ARROW_PRE_0_15_IPC_FORMAT" not in os.environ or os.environ["ARROW_PRE_0_15_IPC_FORMAT"] != 1:
```

(or `os.environ.get("ARROW_PRE_0_15_IPC_FORMAT", 0) != 1`)

src/pandas_profiling/config_minimal.yaml
```diff
@@ -97,6 +97,15 @@ correlations:
       warn_high_correlations: True
       threshold: 0.9
 
+# Spark specific correlations
```
sbrugman, line 100: "Spark-specific configuration"

src/pandas_profiling/config.py
```diff
@@ -166,6 +166,9 @@ def clear(self):
         self.config = confuse.Configuration("PandasProfiling", __name__, read=False)
         self.set_file(str(get_config_default()))
 
+    def set_spark(self):
+        self.set_file(str(get_config("config_default_spark.yaml")))
```
sbrugman, line 170: What is the advantage of creating an additional config file over using `config_default.yaml` for both engines (since the parameters for spark and pandas in the config are split already)? Knowing which engine should be used would be enough.

src/pandas_profiling/model/dataframe_wrappers.py
```diff
@@ -0,0 +1,565 @@
...
+    @abstractmethod
+    def empty(self) -> bool:
+        """
+        return True if dataframe is empty, else return false. A dataframe of NaN should not
+        evaluate to empty
+
+        Returns: True if dataframe is empty else false
+
+        """
+        pass
```
sbrugman, line 102: raise NotImplementedError

src/pandas_profiling/model/dataframe_wrappers.py
```diff
@@ -0,0 +1,565 @@
...
+
+    def __getitem__(self, key):
+        return self.df[key]
+
+    def get_pandas_df(self) -> pd.DataFrame:
+        return self.df
+
+    def head(self, n) -> pd.DataFrame:
+        """
+        TODO - should this return a pd.DataFrame or PandasDataFrame?
```
sbrugman, line 317: pd.DataFrame, there currently is no double usage of the wrapper's result

src/pandas_profiling/profile_report.py
chanedwin: sounds good!

src/pandas_profiling/utils/common.py
chanedwin: thanks! this is way cleaner

src/pandas_profiling/model/typeset.py
chanedwin: changed

src/pandas_profiling/config_minimal.yaml
chanedwin: changed

src/pandas_profiling/config.py
chanedwin: oh i thought you previously didn't want separate parameters for each engine in https://github.com/pandas-profiling/pandas-profiling/pull/670#discussion_r557574342?

chanedwin: ah ok, rereading your comment again, realised I misread it. I think it might be good to have different default values though for each engine (pandas does all by default, spark only does pearson for speed), and that logic has to be stored somewhere. Wondering if the config file a good place to store this? if not it'll have to be embedded in the code.

src/pandas_profiling/model/dataframe_wrappers.py
chanedwin: ok!

src/pandas_profiling/model/dataframe_wrappers.py
chanedwin: ok!

kurosch:
Hi @chanedwin the error is caused by an outdated version of fastparquet. Could you change the requirement in requirements-test.txt to `fastparquet==0.5.0`. Maybe a rebase onto current dev branch would be better?

On a side note @sbrugman how come we use fastparquet when it used to be pyarrow?

sbrugman:
`fastparquet` is only used for testing. We switched from `pyarrow` to [support python 3.8](https://github.com/pandas-profiling/pandas-profiling/commit/036aa91b4ea6437ee0e6190100d2bf25f3a4bed6), which wasn't yet compatible by then. We can switch back and drop the dependency if this no longer causes issues. [Relevant test](https://github.com/pandas-profiling/pandas-profiling/blob/master/tests/issues/test_issue147.py#L16).

PR #680, opened 2021-01-30
Detect dtype from pandas df

mohith7548:
This pull request is to address issue https://github.com/pandas-profiling/pandas-profiling/issues/676 <br>

## Issue
- Developer may choose to change the type of the variable(s) after reading from `pandas`.
- So, the variable type can be changed with the help of `astype` function.
- However, `Pandas-Profiling` is ignores the re-imagined data-type and chooses to infer on its own.
- This is due to the code located here. ( Pointed by @ieaves: https://github.com/pandas-profiling/pandas-profiling/issues/676#issuecomment-767988012 )
https://github.com/pandas-profiling/pandas-profiling/blob/d4e70185fa7e16b1d0b07dfb731897ef06bb3067/src/pandas_profiling/model/summary.py#L43

## Solution
- Suggested by @ieaves here: https://github.com/pandas-profiling/pandas-profiling/issues/676#issuecomment-769209611
- The goal is to not break existing code base & make as minimal required changes as possible.
- Add a new configuration parameter called `infer_dtypes` of `bool` type, which is set `True` by default.
(So as to ensure no code base is effected by this new param)
- `infer_dtypes` is now an optional keyword argument for `ProfileReport` class in ` src/pandas_profiling/profile_report.py` 
- Passing `infer_dtype` variable across many functions is a risky choice. So, it's suggested to use this param in the config file.
- In function `describe_1d` of `src/pandas_profiling/model/summary.py`
    - When `infer_dtypes` is `True` it would run the existing code, i.e infer the `dtype`.
    - When `infer_dtypes` is set `False`, it would not infer the `dtype` rather it detects the `dtype` from the `pandas` dataframe
```python3
    # Get `infer_dtypes` from config
    infer_dtypes = config["infer_dtypes"].get(bool)
    if infer_dtypes:
        # Infer variable types
        vtype = typeset.infer_type(series)
        series = typeset.cast_to_inferred(series)
    else:
        # Detect variable types from `pandas` dataframe
        vtype = typeset.detect_type(series)
```
- Hence `infer_dtypes` is added in `src/pandas_profiling/config_default.yaml` & `src/pandas_profiling/config_minimal.yaml` and is set `True` by default.
- To reflect the same in console version of `pandas_profiling`, two command line arguments are added to match the new bool param.
    - `--infer_dtypes`: equivalent to `infer_dtypes = True` and has no affect anyway since by default this param is set `True`
    - `--no-infer_dtypes`: equivalent to `infer_dtypes = False` and is require when you want `pandas_profiling` to get`dtype` from `pandas` dataframe

## Usage
```python
import numpy as np
import pandas as pd

df = pd.DataFrame({
    'Dummy': ['X', 'Y', 'X', 'X', 'Y', 'Y', 'Y', 'X'],
    'Contract': [9940243658, 9940243537, 9940243103, 9940242844, 9940242844, 9940242840, 9940242774, 9940242774]
})

# New Feature
# Generate report by using the `dtypes` from dataframe. Developer may choose to change the variable type and the changed dtypes are considered.
ProfileReport(df, infer_dtypes=False)
df.profile_report(infer_dtypes=False)

# Current usage 
# Little(No) effect. This generate report by infering dtypes irrespective of change in dtypes by the developer.
ProfileReport(df)
ProfileReport(df, infer_dtypes=True)
df.profile_report()
df.profile_report(infer_dtypes=True)
```

## Additional issue found:
```python3
import numpy as np
import pandas as pd

df = pd.DataFrame({
    'Dummy': ['X', 'Y', 'X', 'X', 'Y', 'Y', 'Y', 'X'],
    'Contract': [9940243658, 9940243537, 9940243103, 9940242844, 9940242844, 9940242840, 9940242774, 9940242774]
})

ProfileReport(df, title='MyAnalysis')    # ---> (1)
ProfileReport(df)                        # ---> (2)
```
- When (1) followed by (2) is executed, the `title` name set in (1) is also appearing in report generated by (2)
- It shouldn't be the case as (2) doesn't have any `title` param (expected title: default title "Pandas Profiling Report")

## Additional issue fix:
- It is assured that by clearing the `config` variable from the cache for every time `ProfileReport` / `profile_report` is called fixes the issue.
- So below line is added as the first line in the constructor of `ProfileReport`, so as to avoid any previous instance of `config` to be used again.
```python
config.clear() # to reset (previous) config.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,9 +160,6 @@ df = pd.DataFrame(
 To generate the report, run:
 ```python
 profile = ProfileReport(df, title="Pandas Profiling Report")
-
-# when dtypes are changed by `astype` function. Detect (new)dtypes from dataframe itself
-profile2 = ProfileReport(df, title="Pandas Profiling Report", infer_dtypes=False)
 ```
 
 ### Explore deeper
@@ -251,14 +248,31 @@ A set of options is available in order to adapt the report generated.
 More settings can be found in the [default configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_default.yaml), [minimal configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_minimal.yaml) and [dark themed configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_dark.yaml).
 
 **Example**
-
+- configure your report. You can refer advanced usage docs [here](https://pandas-profiling.github.io/pandas-profiling/docs/master/rtd/pages/advanced_usage.html)
 ```python
 profile = df.profile_report(title='Pandas Profiling Report', plot={'histogram': {'bins': 8}})
 profile.to_file("output.html")
+```
 
-# If dtypes are changed (using `astype`) and need to consider new dtypes, set `infer_dtypes`: False
-profile2 = df.profile_report(title='Alternative Report', infer_dtypes=False)
-profie2.to_file('output2.html')
+- If dtypes are changed (using `astype`) and need to consider the new dtypes, set `infer_dtypes`: False
+```python3
+df = pd.DataFrame({
+    'Dummy': ['X', 'Y', 'X', 'X', 'Y', 'Y', 'Y', 'X'],
+    'EmpID': [9940243658, 9940243537, 9940243103, 9940242844, 9940242844, 9940242840, 9940242774, 9940242774]
+})
+
+df['EmpID'] = df['EmpID'].astype('str')
+
+# `EmpID` would be considered as `real number` in below report eventhough it is casted as `str` above.
+# This is because by default pandas-profiling tries to infer the dtype.
+profile2 = df.profile_report(title='Inferred dtype - Report')
+profile2.to_file('output2.html')
+
+
+# To read `EmpID` as a category item / use the same dtypes as read by pandas (df.dtypes)
+profile3 = df.profile_report(title='Detected dtype - Report', infer_dtypes=False)
+profile3.to_file('output3.html')
+
 ```
 
 ## Supporting open source
--- a/README.md
+++ b/README.md
@@ -160,6 +160,9 @@ df = pd.DataFrame(
 To generate the report, run:
 ```python
 profile = ProfileReport(df, title="Pandas Profiling Report")
+
+# when dtypes are changed by `astype` function. Detect (new)dtypes from dataframe itself
+profile2 = ProfileReport(df, title="Pandas Profiling Report", infer_dtypes=False)
 ```
 
 ### Explore deeper
@@ -241,6 +244,9 @@ A set of options is available in order to adapt the report generated.
 * `title` (`str`): Title for the report ('Pandas Profiling Report' by default).
 * `pool_size` (`int`): Number of workers in thread pool. When set to zero, it is set to the number of CPUs available (0 by default).
 * `progress_bar` (`bool`): If True, `pandas-profiling` will display a progress bar.
+* `infer_dtypes` (`bool`): 
+  * `True` by default. When `True` the `dtype` of variables are inferred by `pandas-profiling`.
+  * When `False`, the data-types of variables are detected from `pandas`. dataframe. (A variable `dtype` can be changed using `astype` function and the same `dtype` is used by `pandas-profiling`).
 
 More settings can be found in the [default configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_default.yaml), [minimal configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_minimal.yaml) and [dark themed configuration file](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_dark.yaml).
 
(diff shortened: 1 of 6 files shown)


src/pandas_profiling/profile_report.py
```diff
@@ -172,6 +175,12 @@ def description_set(self):
             )
         return self._description_set
 
+    @property
+    def infer_dtypes(self):
```
sbrugman, line 179: This isn't used (and not needed, see line 81)

src/pandas_profiling/profile_report.py
```diff
@@ -90,6 +91,7 @@ def __init__(
         self._json = None
         self._typeset = None
         self._summarizer = None
+        self._infer_dtypes = None
```
sbrugman, line 94: The line above includes infer_dtypes in the settings (`config.set_kwargs(kwargs)`). This line and 151 can be removed

src/pandas_profiling/profile_report.py
```diff
@@ -50,6 +50,7 @@ def __init__(
             sample: optional dict(name="Sample title", caption="Caption", data=pd.DataFrame())
             **kwargs: other arguments, for valid arguments, check the default configuration file.
         """
+        config.clear()  # to reset (previous) config.
```
sbrugman, line 53: Good addition. We should move away from global config someday...

README.md
```diff
@@ -160,6 +160,9 @@ df = pd.DataFrame(
 To generate the report, run:
 ```python
 profile = ProfileReport(df, title="Pandas Profiling Report")
+
+# when dtypes are changed by `astype` function. Detect (new)dtypes from dataframe itself
+profile2 = ProfileReport(df, title="Pandas Profiling Report", infer_dtypes=False)
```
sbrugman, line 165: For a user this comes to early in the documentation. I'd leave this one out (we already have this example below)

README.md
```diff
@@ -249,6 +255,10 @@ More settings can be found in the [default configuration file](https://github.co
 ```python
 profile = df.profile_report(title='Pandas Profiling Report', plot={'histogram': {'bins': 8}})
 profile.to_file("output.html")
+
+# If dtypes are changed (using `astype`) and need to consider new dtypes, set `infer_dtypes`: False
+profile2 = df.profile_report(title='Alternative Report', infer_dtypes=False)
+profie2.to_file('output2.html')
```
sbrugman, line 261: typo `profie2`

README.md
```diff
@@ -241,6 +244,9 @@ A set of options is available in order to adapt the report generated.
 * `title` (`str`): Title for the report ('Pandas Profiling Report' by default).
 * `pool_size` (`int`): Number of workers in thread pool. When set to zero, it is set to the number of CPUs available (0 by default).
 * `progress_bar` (`bool`): If True, `pandas-profiling` will display a progress bar.
+* `infer_dtypes` (`bool`): 
+  * `True` by default. When `True` the `dtype` of variables are inferred by `pandas-profiling`.
```
sbrugman, line 248: Please give an example, users might not directly understand otherwise.

sbrugman:
@mohith7548 Great start, I've added some comments to improve the PR and get it in the next release.

src/pandas_profiling/profile_report.py
ieaves: There's only one way to eat an elephant :).

src/pandas_profiling/profile_report.py
mohith7548: Sorry my bad! I rechecked. Initially I used this method, finally I tried using config.
Yes, we can remove lines that has `infer_dypes` in `profile_report.py` file as the kwarg is stored using `config.set_kwargs(kwargs)`.

src/pandas_profiling/profile_report.py
mohith7548: Yes, not needed.
We can remove this function as well, including the lines which has `infer_dtypes` word

README.md
mohith7548: sure.

README.md
mohith7548: Actually it's complicated to show in the doc. Maybe we can include in detail with example in [adv docs](https://pandas-profiling.github.io/pandas-profiling/docs/master/rtd/pages/advanced_usage.html)

ieaves:
This is great! I actually don't think it would be difficult to use a configuration setting on the ProfileReport object rather than the global config here though. The current complication is passing those configurations down to describe_1d through so many layers of the call stack (`ProfileReport.description_set` -> `describe_df` -> `get_series_descriptions` -> `describe_1d`). `ProfileReport.description_set` is just a thin wrapper around the `describe_df`  with `get_series_description` as the first place it switches from operating over dataframes to series. If we move some of this logic up into either [describe](https://github.com/pandas-profiling/pandas-profiling/blob/4dd35dfbf7c1eb005a90e87275efdeec6968931b/src/pandas_profiling/model/describe.py#L24) or [get_series_descriptions](https://github.com/pandas-profiling/pandas-profiling/blob/4dd35dfbf7c1eb005a90e87275efdeec6968931b/src/pandas_profiling/model/summary.py#L68) are targets further up the stack we might be interested in using. 

In the short term, we could simply pass self from the ProfileReport down into `describe_df` and use the `infer_dtypes` attribute defined here to split out inference vs detection. In the long term, this could evolve into something like a `ProfileReportConfig` or something which gets passed around instead of the global config state. Not sure what you think @sbrugman?

sbrugman:
@ieaves For the PR I'd keep the global config and start on migrating to a solution similar as you described in it's own PR.  Ditching the global config is on the project's wishlist (there are other issues, for instance that `confuse` does not handle comments adequately), however it's hard to find the time. Hoping that someone is able to contribute it one day.

README.md
mohith7548: Okay, let me try

sbrugman:
@mohith7548 Thanks for your effort on this. I've processed the remaining comments and will merge in that PR.

PR #691, opened 2021-02-08
Consolidated the to_expectation_suite methods

talagluck:
After some discussion with @spbail, we though it made sense to have just one `to_expectation_suite` method, with parameters for saving suites, running validations, and building data docs. We combined them and updated the examples to match.

### DIFF
--- a/examples/features/great_expectations_example.py
+++ b/examples/features/great_expectations_example.py
@@ -1,8 +1,9 @@
+import great_expectations as ge
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


sbrugman:
PR looks good, thanks. The formatting is automated @talagluck, you can run it using `make lint` or `pre-commit run --all-files`, which will make the CI/CD checks to pass.

lgtm's alerts are false positives in the example, can be ignored.

talagluck:
Thanks, Simon! Just updated the linting.

PR #696, opened 2021-02-12
feat: add number and prct of negative values

### LINKED ISSUE #695: Counting negative values
**Missing functionality**
When asked to provide data overview, the number of negative value (and percentage) can be really helpful to assess data quality. I would like to add the number and the percentage of negative values to the summary.

**Proposed feature**
Look like it could be an easy implementation. Something like 
summary['n_neg'] = series.lt(0).sum()
summary['p_neg'] = series.lt(0).mean()

in describe_numeric_1d

**Alternatives considered**
None

**Additional context**
Would be happy to open my first ever  PR

gverbock:
Add the count of negative values and the associated percentage to the statistics. Closes #695


sbrugman:
The proposed implementation is inefficient:
-  "p_neg" can be computed from "n_neg" and "n"
- `series.lt(0)` needs to process the full series once, however we already do that during the value counts. Computing the negative values of the counts will give an enormeous perofmance gain in most cases (see `n_zeros` as example on how to do this).

Other comments:
- This functionality need to be unit tested;
- This functionality may add a configuration parameter

sbrugman (changes requested):
See comment in PR

gverbock:
Thanks for the comments. Will work on it.

sbrugman:
Looks good.

The statistics are now computed and show up in the internal representation. We should also include them in the report.  We can simply copy the part for `n_zeros` and `p_zeros` [here](https://github.com/pandas-profiling/pandas-profiling/blob/develop/src/pandas_profiling/report/structure/variables/render_real.py#L106) and move mean to the other column (to keep them equal length).

image

Another thing: you can add the change to the changelog of the next version (`docsrc\source\pages\changelog\2_12_0.rst`). You can put a summary, link to the issue and your github username (e.g. "contributed by @gverbock")

Lastly, the commits are not linked to your github user but name. This is [easy to fix](https://stackoverflow.com/questions/1746562/github-shows-name-instead-of-a-linked-username-in-commits).

gverbock:
I did the following: 

- Put the statistic on the report.
- Add the log.

I see that the statistics are now showing up like this. 

image


Are you happy with the lay-out or would you like to change it?

PR #697, opened 2021-02-16
Added expect_column_values_to_be_between for numeric and datetime columns.

talagluck:
These expectations seemed straightforward, so I added them to the expectation_algorithms.

### DIFF
--- a/src/pandas_profiling/model/expectation_algorithms.py
+++ b/src/pandas_profiling/model/expectation_algorithms.py
@@ -36,6 +36,11 @@ def numeric_expectations(name, summary, batch, *args):
             name, strictly=summary["monotonic_decrease_strict"]
         )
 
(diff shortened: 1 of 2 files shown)


src/pandas_profiling/model/expectation_algorithms.py
```diff
@@ -36,6 +36,11 @@ def numeric_expectations(name, summary, batch, *args):
             name, strictly=summary["monotonic_decrease_strict"]
         )
 
+    if summary.get("min") or summary.get("max"):
```
sbrugman, line 39: Strictly `get` should be used for obtaining the value, not for membership checks. 

(better to use `in`: `if any(k in summary for k in ["min", "max"])`)

sbrugman:
LGTM. Only thing missing is to add the expectations to the docs page.

Issue #678, opened 2021-01-28
Feature request: df can be empty


jamesmyatt:
**Missing functionality**

When I run the profiler with an empty datafram---in this case, there are no rows but some columns---I get a ValueError with "df can not be empty".

```{python}
if df.empty:
   raise ValueError("df can not be empty")
```

This seems like an unnecessary limitation. When I'm profiling a relational database, then I want to create a profile report for every table in the database, including the empty ones.

**Proposed feature**

Allow empty data frames.

**Alternatives considered**

N/A

**Additional context**

#664 seems relevant too.

sbrugman:
The linked issue is unrelated.

This should be a simple way to get started with contributing to pandas-profiling/open source. If anyone is interested in taking this up, please send in a PR to get started or ask here or on our slack channel.

jamesmyatt:
Thanks

PR #703, opened 2021-02-17
feat: Allow empty dataframe

fwd2020-c:
Fixes issue #678 
Work in progress, need to modify tests.
Does this approach look good to you?
Worked on this with @spbail 😄

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       additional_dependencies: [ pyupgrade==2.7.3 ]
       args: [ --nbqa-mutate, --py36-plus ]
 -   repo: https://github.com/asottile/pyupgrade
-    rev: v2.9.0
+    rev: v2.10.0
     hooks:
     -   id: pyupgrade
         args: ['--py36-plus','--exit-zero-even-if-changed']
--- a/README.md
+++ b/README.md
@@ -238,11 +238,13 @@ A set of options is available in order to adapt the report generated.
 * `title` (`str`): Title for the report ('Pandas Profiling Report' by default).
 * `pool_size` (`int`): Number of workers in thread pool. When set to zero, it is set to the number of CPUs available (0 by default).
 * `progress_bar` (`bool`): If True, `pandas-profiling` will display a progress bar.
+* `infer_dtypes` (`bool`): When `True` (default) the `dtype` of variables are inferred using `visions` using the typeset logic (for instance a column that has integers stored as string will be analyzed as if being numeric).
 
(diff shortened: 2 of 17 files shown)


sbrugman:
Really good that you're starting contributing to open source :) Feel free to ask if you're unsure about certain parts of the change or contribution process.

A few notes:
- [test] As you mention, it would be good to have a dedicated test for `n=0` (suggestion to add unit tests here: https://github.com/pandas-profiling/pandas-profiling/blob/develop/tests/unit/test_summary_algos.py)
- [test] We would need another test to see if the rendered HTML makes sense.
- [code] The `if df.empty` check can be removed entirely.

fwd2020-c:
Hi Simon,
Thank you so much for your feedback and I followed your instructions to make some changes:

1. [test] As you mention, it would be good to have a dedicated test for n=0 (suggestion to add unit tests here: https://github.com/pandas-profiling/pandas-profiling/blob/develop/tests/unit/test_summary_algos.py)
2.  [code] The if df.empty check can be removed entirely.

However, I have no idea how to create a test to check rendered HTML. 

What do you think?

sbrugman:
Making progress!

The unit tests should add cases for these functions (all affected by removing the exception):
- `get_table_stats` (empty DataFrame)
- `describe_counts`, `describe_generic` and `describe_supported` (empty Series)

I haven't tried running the HTML or Jupyter notebook widgets yet myself, but expect there are quite some parts that should be hidden/empty for empty DFs. We can explicitly test for them in the HTML/Notebooks. If you'd like some feedback, you can either upload a screenshot or the report for an empty dataframe here.

Another thing: you can add the change to the changelog of the next version (`docsrc\source\pages\changelog\2_12_0.rst`). You can put a summary, link to the issue and your github username (e.g. "contributed by @fwd2020-c")

fwd2020-c:
Simon, I created the tests for `describe_counts` , `describe_generic`, `describe_supported`. They passed pytest but I am not very sure it is the right way to test describe_supported which is wrapped by a decorator.

`get_table_stats` needs arg of variable_stats, which is ,I guess, the output of `describe_1d`:
`describe_1d(df, summarizer: BaseSummarizer, typeset: Unsupported)` 
I got an attribute error of "no attribute 'infer_type'"
Is it ok to provide me an example on how to solve this one? 

Attached html report:
http://localhost:8888/view/empty_df_report.html

sbrugman:
`variable_stats` is just a dictionary, you could start with an empty dict and add the testing keys needed.

You've shared a link to the file that only you can access. It should be possible to drag the html file to the comment section and github will upload it for you.

fwd2020-c:
Simon, thank you so much! empty dictionary works! (Out of curiosity, how come empty dictionary can replace the arg of variable_stats: Previously calculated statistic on the DataFrame? 🤔)

http://localhost:8888/view/dragged_empty_df_report.html

PR #704, opened 2021-02-18
modified summarizer

ieaves:


src/pandas_profiling/model/handler.py
```diff
@@ -40,6 +51,9 @@ def _complete_dag(self):
         ):
             self.mapping[to_type] = self.mapping[from_type] + self.mapping[to_type]
 
+    def op(self, dtype: Type[VisionsBaseType]) -> Callable:
+        return compose(self.mapping.get(dtype, []))
```
sbrugman, line 55: Any reason not to replace `op` in handle with `self.op`?

ieaves: Ahh, none in particular, the op method was actually old code we could alternatively remove.

PR #708, opened 2021-02-19
build: Move ipywidgets to extras

akx:
This will allow installing the package without installing all of Jupyter, if you only want HTML output.

There's a PR over at PhiK's repo to get rid of the numba/llvmlite hard dependency; an earlier revision of this PR also took care of making PhiK optional. https://github.com/KaveIO/PhiK/pull/17

### DIFF
--- a/docsrc/source/pages/changelog/v2_12_0.rst
+++ b/docsrc/source/pages/changelog/v2_12_0.rst
@@ -27,4 +27,4 @@ Changelog v2.12.0
 
 ⬆️ Dependencies
 ^^^^^^^^^^^^^^^^^^
--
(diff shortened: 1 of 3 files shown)


sbrugman:
Thanks @akx. 

I think phik is valueable to be installed by default, we should loosen the requirements there.
The changelog has a specific section for dependencies (`docsrc\source\pages\changelog\v2_12_0.rst`)

akx:
> I think phik is valueable to be installed by default, we should loosen the requirements there.

Waiting for them to respond in the related PR, then.

> The changelog has a specific section for dependencies (`docsrc\source\pages\changelog\v2_12_0.rst`)

Added a note about this there.

Issue #538, opened 2020-08-06
Start Matplotlib during the to_html method


Shihao-Zhong:
**Describe the bug**
I am trying to use the pandas-profiling to build a backend service that generate some static report for some of the data in our system. When I trying to use the to_html method to generate report for a dataset, I can see a Matplotlib UI is started in backend, it's there anyway to generate the report that does not trigger a Matplotlib application? I am afraid that there may have some performance issues for my flask application.

**To Reproduce**

_Code:_ Preferably, use this code format:
```python
"""
Test for issue XXX:
https://github.com/pandas-profiling/pandas-profiling/issues/XXX
"""
import pandas as pd
import pandas_profiling

profile = ProfileReport(df, title=report_title, explorative=True, minimal=True)
html = profile.html()

```

**Version information:**

2.8.0

**Additional context**

sbrugman:
Good question, please see these answers: 

https://github.com/pandas-profiling/pandas-profiling/issues/373#issuecomment-586579944
https://stackoverflow.com/a/4935945/470433

Issue #373, opened 2020-02-10
While calling through rest api : RuntimeError: main thread is not in main loop


rushikesh79:
I tried with both version of pandas profiling 2.3 and 2.4
When I try to hit the api for profiling it generate a html file correactly, when i hit that api for 2nd time I got an error:
Exception ignored in: <function Image.__del__ at 0x0D771C48>
Traceback (most recent call last):
  File "C:\Users\RS54863\AppData\Local\Programs\Python\Python37-32\lib\tkinter\__init__.py", line 3507, in __del__
    self.tk.call('image', 'delete', self.name)
RuntimeError: main thread is not in main loop
Exception ignored in: <function Image.__del__ at 0x0D771C48>
Traceback (most recent call last):
  File "C:\Users\RS54863\AppData\Local\Programs\Python\Python37-32\lib\tkinter\__init__.py", line 3507, in __del__
    self.tk.call('image', 'delete', self.name)
RuntimeError: main thread is not in main loop
Exception ignored in: <function Image.__del__ at 0x0D771C48>
Traceback (most recent call last):
  File "C:\Users\RS54863\AppData\Local\Programs\Python\Python37-32\lib\tkinter\__init__.py", line 3507, in __del__
    self.tk.call('image', 'delete', self.name)
RuntimeError: main thread is not in main loop
Exception ignored in: <function Image.__del__ at 0x0D771C48>
Traceback (most recent call last):
  File "C:\Users\RS54863\AppData\Local\Programs\Python\Python37-32\lib\tkinter\__init__.py", line 3507, in __del__
    self.tk.call('image', 'delete', self.name)
RuntimeError: main thread is not in main loop
Tcl_AsyncDelete: async handler deleted by the wrong thread

rushikesh79:
I tried with both version of pandas profiling 2.3 and 2.4

sbrugman:
Could you provide the minimal information to reproduce this error?

- the minimal code you are using to generate the report
- which environment you are using (jupyter notebook, console or IDE) and packages (`pip freeze > packages.txt`)
- a sample or description of the dataset (`df.head()`, `df.info()`)

rushikesh79:
def generateAnalysisReport(df, fileReference):
    try:
        result = {}
        profile = df.profile_report(title='Pandas Profiling Report')
        reportFileName= fileReference.split('.pkl')[0] + ".html"
        reportFilePath= REPORTS_FOLDER_PATH + reportFileName
[...]

PR #798, opened 2021-06-08
matplotlib Agg

endremborza:
one proposed way to handle #373 and #538

as I don't have a better idea but I don't really like people proposing quick fixes without attempting them themselves, here it is


endremborza:
its rather annoying that for some reason gitlub actions does not show output of why it is failing, so I'm sorry but I have no idea whats breaking

sbrugman:
The errors aren't coming from your PR. Thanks !

PR #818, opened 2021-08-14
Implemented simple PDF generation through weasyprint

kurosch:
Added the possibility to generate PDFs from the HTML reports through weasyprint. 

 **TODO**:
- [ ] add CSS format for prettier reports

### DIFF
--- a/examples/census/census.py
+++ b/examples/census/census.py
@@ -41,7 +41,7 @@ if __name__ == "__main__":
     df = df.replace("\\?", np.nan, regex=True)
 
     # Initialize the report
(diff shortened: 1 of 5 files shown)


sbrugman:
@kurosch Thanks for the contribution! 

Introducing code and a dependency increases complexity and reduces maintainability. At least the dependency should be optional. It's much more logical to apply the conversion step after generating the report. The html -> pdf functionality itself could be an extension.

With the current style I also expect the PDF output to not be useful, for the reason that you already mention (CSS not being optimized for static output). However, If you're planning on optimizing the CSS for print then that would be an amazing contribution.

PR #875, opened 2021-11-06
initial tests for describe

chanedwin:
This commit adds the initial batch of tests for the describe function so that describe is tested by default


tests/backends/spark_backend/test_descriptions_spark.py
```diff
@@ -0,0 +1,439 @@
...
+    cfg.missing_diagrams["bar"] = False
+    cfg.missing_diagrams["dendrogram"] = False
+    cfg.missing_diagrams["heatmap"] = False
+    cfg.missing_diagrams["matrix"] = False
+
+    # cfg["spark"]["quantile_error"].set(0)
+
+    spark = spark_session
+
+    if column == "mixed":
```
sbrugman, line 403: This preprocessing should not be here. Either the profiler supports it, or not and errors should be raises.

sbrugman, line 403: This preprocessing should not be here. Either the profiler supports it, or not and errors should be raises.

sbrugman:
The CI seems to fail  because pyspark  is not installed.

sbrugman:
The CI seems to fail  because pyspark  is not installed.

Issue #863, opened 2021-10-20
Feature: Stop words


gustavorps:
**Missing functionality**
Able to pass a stop words list to improve Words tab details

**Proposed feature**
```py
profile.config.vars.cat.stop_words = [" ", "  ", "the", "and", "at"]
```

[`config.yaml`](https://github.com/pandas-profiling/pandas-profiling/blob/master/src/pandas_profiling/config_default.yaml)
```yaml
...
vars:
    ...
    cat:
        ...
        stop_words: [" ", "  ", "the", "and", "at"]
    ...
...
```

**Additional context**
![image](https://user-images.githubusercontent.com/516827/138155030-3dbdb283-2c0a-4790-8d4d-56b214ad9b3f.png)

rluthi:
Hi, I'm happy take this one. This will be my first contribution

rluthi:
This issue has been resolved in PR #877 which has been merged into the develop branch

gustavorps:
Thanks @rluthi and @sbrugman

PR #877, opened 2021-11-10
ENH:  issue #863 If defined, stop words to be ignored in word count table 

rluthi:
Stop words can be defined by changing the config of the ProfileReport:
-> profile.config.vars.cat.stop_words = [the, to]

The stop words are then ignored in the word count and frequency table : 

image

My changes have no impact on the character count and frequency table.

### DIFF
--- a/src/pandas_profiling/model/pandas/describe_categorical_pandas.py
+++ b/src/pandas_profiling/model/pandas/describe_categorical_pandas.py
@@ -1,7 +1,7 @@
 import contextlib
 import string
 from collections import Counter
-from typing import Tuple
+from typing import List, Tuple
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


src/pandas_profiling/model/pandas/describe_categorical_pandas.py
```diff
@@ -147,6 +160,10 @@ def word_summary_vc(vc: pd.Series) -> dict:
     word_counts = word_counts[word_counts.index.notnull()]
     word_counts = word_counts.groupby(level=0, sort=False).sum()
     word_counts = word_counts.sort_values(ascending=False)
+
+    # Remove stop words
+    word_counts = word_counts.loc[~word_counts.index.isin(stop_words)]
```
sbrugman, line 165: Perhaps only do  this when `len(stop_words) > 0`

sbrugman:
On the right track, please add a unit test for `word_summary_vc`

src/pandas_profiling/model/pandas/describe_categorical_pandas.py
rluthi: > On the right track, please add a unit test for `word_summary_vc`.

Where is the best place to add this unit test ?

PR #891, opened 2021-12-01
feat(spark): enable default spark settings

chanedwin:
Create default settings for spark and
enable automatic switching in
ProfileReport based on input df type

### DIFF
--- a/src/pandas_profiling/config.py
+++ b/src/pandas_profiling/config.py
@@ -303,6 +303,46 @@ class Settings(BaseSettings):
         return self.parse_obj(self.copy(update=update))
 
 
+class PandasSettings(Settings):
+    pass
+
+
+class SparkSettings(Settings):
+    # TO-DO write description
+    vars: Univariate = Univariate()
+
+    vars.num.low_categorical_threshold = 0
+
+    infer_dtypes = False
+
+    correlations: Dict[str, Correlation] = {
+        "spearman": Correlation(key="spearman"),
+        "pearson": Correlation(key="pearson"),
+        "kendall": Correlation(key="kendall"),
+        "cramers": Correlation(key="cramers"),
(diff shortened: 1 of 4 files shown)


src/pandas_profiling/config.py
```diff
@@ -303,6 +303,46 @@ def update(self, updates: dict) -> "Settings":
         return self.parse_obj(self.copy(update=update))
 
 
+class PandasSettings(Settings):
+    pass
+
```
chanedwin, line 308: this was done to ensure consistency. Settings is the "default settings", and engine-specific settings override certain parts of those default settings. All usages of Settings should be switched to PandasSettings (Not done in this PR)

src/pandas_profiling/profile_report.py
```diff
@@ -77,7 +77,9 @@ def __init__(
         if df is None and not lazy:
             raise ValueError("Can init a not-lazy ProfileReport with no DataFrame")
 
-        report_config: Settings = Settings() if config is None else config
+        report_config: Settings = (
+            self.get_default_settings(df) if config is None else config
+        )
```
chanedwin, line 82: here the code calls a function that gets the default settings based on DF type

src/pandas_profiling/profile_report.py
```diff
@@ -424,3 +426,9 @@ def _repr_html_(self) -> None:
     def __repr__(self) -> str:
         """Override so that Jupyter Notebook does not print the object."""
         return ""
+
+    def get_default_settings(self, df) -> Settings:
+        if isinstance(df, pd.DataFrame) or isinstance(df, pd.Series):
```
sbrugman, line 431: isinstance's second argument can be a tuple:
`if isinstance(df, (pd.DataFrame, pd.Series))`

PR #896, opened 2021-12-14
docs: fix typo in link

CodeCox:
The current URL link results in an `404 - File Not Found` error.

This is because the filename has a `.html` extension instead of `.py`.

However, just correcting the extension will make the file downloadable which is not the desired outcome. So I have changed the full path so that the example code opens in github like other source examples.


sbrugman:
Thanks for your contribution!

PR #903, opened 2022-01-02
fix: fix skewness and kurtosis errors

chanedwin:
Spark kurtosis and skewness functions changed
the output from np.nan to None, which caused
some errors. Using pd.isnull instead of np.isnan
enables the handling of both None and np.nan
better. Also make kurtosis and skewness fields
optional.


mkarbo:
This fixes errors in general, not just for spark - I was about to open a PR with this fix as well.

PR #930, opened 2022-02-25
feat: show cat freq with stacked barh not pie - fixes #679

### LINKED ISSUE #679: Feature request: Replace pie chart with alternative
**Missing functionality**

Pie charts are horrible. 😝 

**Proposed feature**

Can we have an option to disable the pie chart and/or to replace it with another format, such as a tree map, please?

**Alternatives considered**

 N/A

**Additional context**

Related issues: https://github.com/pandas-profiling/pandas-profiling/issues?q=pie+chart

rluthi:
Replaces the pie plot by a **horizontal stacked bar chart** to represent category frequency. It is used for boolean and categorical features (which have less than _n_max_ unique values.

Example of the new **horizontal stacked bar chart** representation:
<kbd>

</kbd>

The old pie chart can still generate by setting `profile.config.plot.cat_freq.type == "pie"`:
<kbd>

</kbd>

The  `CatFrequencyPlot`, has the same setting as the former `PiePlot` 
-    `colors`
-    `max_unique`

Additionally, it has a boolean `show` setting to turn this chart on and off entirely. 

The `profile.config.plot.pie` setting is removed. This breaks compatibility but is a cleaner implementation that would allow for other category frequency plots to be implemented in the future  (i.e. treemap)


src/pandas_profiling/config_default.yaml
```diff
@@ -135,8 +135,9 @@ plot:
         # https://github.com/ResidentMario/missingno/issues/93#issuecomment-513322615
         force_labels: true
 
-    pie:
-        # display a pie chart if the number of distinct values is smaller or equal (set to 0 to disable)
+    cat_frequency:
```
a-chernov, line 138: Should it be `cat_freq`?

src/pandas_profiling/config_minimal.yaml
```diff
@@ -137,10 +137,11 @@ plot:
         # https://github.com/ResidentMario/missingno/issues/93#issuecomment-513322615
         force_labels: true
 
-    pie:
-      # display a pie chart if the number of distinct values is smaller or equal (set to 0 to disable)
-      max_unique: 0
-      colors: null # use null for default or give a list of matplotlib recognised strings
+    cat_frequency:
```
a-chernov, line 140: Should it be `cat_freq`?

a-chernov:
Displaying data on chart legend doesn't match w/ `sensitive=True` logic

a-chernov:
@rluthi @sbrugman please see my comments above, thanks

PR #980, opened 2022-05-13
meteorite example: add dependency workaround

stuart-mclaren:
The meteorite example is currently not working.

It gives the following error:

```
cannot import name 'soft_unicode' from 'markupsafe'
```

This seems to be due to markupsafe removing soft_unicode:
https://github.com/pallets/markupsafe/issues/304

We can work around this by pinning to an earlier version of markupsafe.

This gives the following warning:

pandas-profiling 3.2.0 requires markupsafe~=2.1.1, but you have markupsafe 2.0.1 which is incompatible.

but the example seems to work ok.

### DIFF
--- a/docsrc/source/pages/installation.rst
+++ b/docsrc/source/pages/installation.rst
@@ -17,14 +17,14 @@ Using pip
   :alt: PyPi Version
   :target: https://pypi.org/project/pandas-profiling/
 
-You can install using the pip package manager by running
+You can install using the ``pip`` package manager by running:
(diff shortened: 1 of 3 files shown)


sbrugman:
Hi @stuart-mclaren, thanks for flagging this. The issue has been resolved [here](https://github.com/ydataai/pandas-profiling/commit/7653f531aae2b34af96fcf6a97908a18535fd83f) and will be included in the next release.

PR #982, opened 2022-05-16
Deprecate pdoc-based documentation page

dr-ydata:
Sphinx with an RTD theme is the current documentation system. However, a pdoc3 -based documentation page is still up, although it's not being used. This branch deprecates it:

- Removes all `pdoc` related instructions from makefiles and similars.
- Updates links to documentation across the repo (from `docs/rtd/` to `/docs/`)

**Still missing:** 
- Set up redirect from `/docs/rtd` to `/docs/` (release CI/CD pipeline)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report_form.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report_form.yaml
@@ -9,7 +9,7 @@ body:
       value: |
         Thanks for taking the time to fill out this bug report!
         Provide a concise summary of the issue in the title field above.
-        The [documentation](https://pandas-profiling.ydata.ai/docs/master/rtd/pages/support.html) contains instructions on creating bug reports.
+        The [documentation](https://pandas-profiling.ydata.ai/docs/master/pages/support.html) contains instructions on creating bug reports.
   - type: textarea
     id: current-behavior
     attributes:
@@ -62,7 +62,7 @@ body:
     id: dependencies
     attributes:
       label: Dependencies
-      description: Please report versions of relevant dependencies (e.g. via `pip freeze` or `conda list`, [more information](https://pandas-profiling.ydata.ai/docs/master/rtd/pages/support.html))
+      description: Please report versions of relevant dependencies (e.g. via `pip freeze` or `conda list`, [more information](https://pandas-profiling.ydata.ai/docs/master/pages/support.html))
       placeholder: |
         pandas==1.4.2
         numpy==2.3.4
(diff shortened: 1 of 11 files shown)


dr-ydata:
> > @sbrugman Would any of the solutions here for the redirect be a viable option? https://superdevresources.com/redirects-jekyll-github-pages/
> > For the [docs index page](https://github.com/ydataai/pandas-profiling/blob/gh-pages/index.html), a redirect of that kind is used.
> 
> The challenge is not having to create individual redirects for each old page, while not having access to the normal server-side tools. We could create a 404 page that, when the URL contains /rtd/, redirects to the parent folder. e.g. https://pandas-profiling.ydata.ai/docs/master/rtd/pages/previously_existing_page.html See [this post](https://stackoverflow.com/a/65533465) and the [github docs](https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site)

I was checking with YData's infrastructure team and it's not easy to create server-side redirects given the way the website is currently set up. So, the two options I see are: 

- The 404 page method you suggest.
- Suggested by @portellaa, the Sphinx extension [sphinx-reredirects](https://documatt.gitlab.io/sphinx-reredirects/usage.html), which might just work via its wildcard support. 

Both seem fine to me. Assuming it works, the extension would probably lead to a smoother experience. Any preference?

sbrugman:
Sphinx-redirect is preferable, as it can be maintained without modifying the GitHub Pages config. It supports directory redirects.

dr-ydata:
@sbrugman Briefly played around with the extension and it doesn't exactly fit this use case without some convoluted workflows and poor generalisation if/when new pages are added (due to the way the wildcard and source matching works). 

Given that all the links in the repo have been updated (apart from those eventually present in GitHub issues), the problem would mainly manifest with external links from the web. I feel it's OK to redirect those to the main page (and hopefully the new docs structure will make it easy for everyone to find what they are looking for). 

In that case, from what I read on that StackOverflow page, I think a simple solution would be serving a `404.html` page redirecting to the documentation's home. I think this could be achieved, in the `release.yml` CI pipeline, through a:

```bash
cp docs/${{ env.BRANCH_NAME }}/index.html 404.html
git add 404.html
```

Thoughts?

sbrugman:
Ok, let's go with the redirect. By default we can redirect urls that contain `/rtd/` to the same url where the directory is removed (using javascript) to ease the user experience for external web links.

dr-ydata:
@sbrugman As we agreed, given the incoming docs restructuring branch, a lot of external URLs will be broken, it doesn't make as much sense as before to do the `rtd/` redirect, so we will go with the default 404 page redirecting to the docs' home. 

In that case, I am marking this branch as ready for review and creating a separate one with the 404 to merge into `gh-pages`.

Issue #983, opened 2022-05-17
ProfileReport.to_json() returns invalid JSON


LukasBoersma:
### Current Behaviour

Python has the known issue that its standard json implementation [produces invalid JSON](https://bugs.python.org/issue40633): NaN and Infinity values are not supported by the json specification, but Python serializes those values anyway.

Because of that, when using `ProfileReport.to_json()`, the JSON is often not valid. For example, the example code below produces this output that other JSON implementations will fail to parse:

```
...
"variance": 0.0,
"kurtosis": NaN,
"skewness": 0,
...
```

I did this on the `spark-branch` , commit `9017c4a5e26e22152ed3f24c5ec628f70859fa14`

### Expected Behaviour

This is not really the fault of pandas-profiling, but it would be really nice if `to_json` could return valid JSON.

I see several ways to solve this:

 - Never actually return any NaNs in the produced statistics, always set values to None. This would have the problem that some parts of the report contain actual input data and that could still contain NaNs
 - Use a different JSON library that produces a valid output
 - (Maybe the easiest solution) Check for non-finite numbers in the existing `encode_it` function and replace them with None so that the JSON will contain null values

The changed encode_it function could look like this (I would be happy to send a pull request if that is the accepted solution):

```python
def encode_it(o: Any) -> Any:
  if isinstance(o, dict):
      return {encode_it(k): encode_it(v) for k, v in o.items()}
  else:
      if isinstance(o, (bool, int, str)):
          return o
      if isinstance(o, float):
          if not math.isfinite(o):
              # Encode non-finite floats as None.
              # This is necessary because JSON does not support NaN/Infinity values.
              return None
          else:
              return o
      elif isinstance(o, list):
          [...]
```

### Data Description

`DataFrame([1, 1, 1], columns=["a"])`

### Code that reproduces the bug

```Python
import pandas as pd
from pandas_profiling import ProfileReport

df = pd.DataFrame([1, 1, 1], columns=["a"])

profile = ProfileReport(df, title="Pandas Profiling Report", minimal=True)

print(profile.to_json())
```

### pandas-profiling version

spark-branch @ 9017c4a5e26e22152ed3f24c5ec628f70859fa14

### Dependencies

```Text
joblib~=1.1.0
scipy>=1.4.1
matplotlib>=3.2.0
pydantic>=1.8.1
PyYAML>=5.0.0
jinja2>=2.11.1
markupsafe~=2.0.1
visions[type_image_path]==0.7.4
[...]

PR #996, opened 2022-06-08
JSON encode of non-finite floats

LukasBoersma:
As described in #983, the JSON output is currently generating invalid JSON when non-finite floats (like NaN or Infinity) are included.

As @sbrugman suggested, I added a config option to switch between three behaviours:

- `PYTHON_NATIVE`: Use Python's default behaviour and generate the invalid JSON (as before)
- `NULL`: Encode non-finite numbers as null values.
- `STRING`: Stringify non-finite numbers (for example, NaN becomes the string "nan")

I made the null output the new default because it seems less surprising to encounter a null value when parsing numbers in the JSON compared to a  "nan" string. I see that there is also an argument to make the old behaviour the default, to ensure 100% compatibility with existing code. Please let me know what you think should be the default.

## Summary of the changes

- I added an enum `JsonNonFiniteEncoding` with the three options listed above
- I added an `json_non_finite_encoding` field in the `Settings` class
- In the `encode_it` method that already manages the JSON-encoding of all values, I added special handling for non-finite floats, which encodes the values according to the configuration.

The changes should not break anything except that the JSON output will now contain null values instead of the raw "NaN" entries that violate the standard.

## Example usage

```python
import pandas as pd
from pandas_profiling.config import Settings, JsonNonFiniteEncoding

df = pd.DataFrame([1, 1, 1], columns=["a"])

profile = ProfileReport(df, title="Pandas Profiling Report", minimal=True)

profile.config.json_non_finite_encoding = JsonNonFiniteEncoding.STRING

print(profile.to_json())
```

Gives the following output:

```
[...]
"kurtosis": "nan",
[...]
```

When choosing the `NULL` config option, it would be:

```
"kurtosis": null,
```

And with the `PYTHON_NATIVE` option it would be:

```
"kurtosis": NaN,
```

### DIFF
--- a/src/pandas_profiling/config.py
+++ b/src/pandas_profiling/config.py
@@ -180,6 +180,16 @@ class Html(BaseModel):
     full_width: bool = False
 
 
+# Describes how non-finite numbers (such as NaN or Infinity) should be encoded in the JSON output
+class JsonNonFiniteEncoding(Enum):
+    # Use the default python behaviour, which violates the official JSON standard
+    PYTHON_NATIVE = 0
+    # Encode non-finite numbers as null values
+    NULL = 1
+    # Encode non-finite numbers as null values
+    STRING = 2
+
+
 class Duplicates(BaseModel):
     head: int = 10
     key: str = "# duplicates"
@@ -292,6 +302,9 @@ class Settings(BaseSettings):
     n_obs_unique: int = 10
     n_freq_table_max: int = 10
     n_extreme_obs: int = 10
+
+    # JSON output
+    json_non_finite_encoding: JsonNonFiniteEncoding = JsonNonFiniteEncoding.NULL
 
     # Report rendering
     report: Report = Report()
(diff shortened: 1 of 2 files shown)


src/pandas_profiling/config.py
```diff
@@ -180,6 +180,16 @@ class Html(BaseModel):
     full_width: bool = False
 
 
+# Describes how non-finite numbers (such as NaN or Infinity) should be encoded in the JSON output
+class JsonNonFiniteEncoding(Enum):
+    # Use the default python behaviour, which violates the official JSON standard
+    PYTHON_NATIVE = 0
+    # Encode non-finite numbers as null values
+    NULL = 1
+    # Encode non-finite numbers as null values
```
alexbarros, line 189: the same comment, can you update the string one?

alexbarros (commented):
Other than the commitlint issues looks good to me, can you fix it so we can merge this?
Also, can you add some tests?

chanedwin:
@LukasBoersma could you change this PR into develop instead of spark-branch? it isn't a spark specific issue

PR #1011, opened 2022-07-14
Nf floats

Vamp1899:
This issue is the same issue as mentioned by Lukas but there's a change in code structure by keeping privacy of variables in consideration with regards to more upcoming development .

Pointers - 
Handles NaN values in Json as
Python native type  - which generates error
Float Null type (Nan , infinity) - Passed as a string in final JSON
Null type - Which modifies values as null in JSON more like allow_nan = True in json python package

### DIFF
--- a/src/pandas_profiling/config.py
+++ b/src/pandas_profiling/config.py
@@ -1,9 +1,13 @@
 """Configuration for the package."""
 from enum import Enum
 from typing import Any, Dict, List, Optional
-
+import warnings
 from pydantic import BaseModel, BaseSettings, Field
 
+# Comment this function to see Warnings in console
+def warn(*args, **kwargs):
+    pass
+warnings.warn = warn
 
 def _merge_dictionaries(dict1: dict, dict2: dict) -> dict:
     """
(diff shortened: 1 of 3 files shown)


src/pandas_profiling/config.py
```diff
@@ -1,9 +1,13 @@
...
 from enum import Enum
 from typing import Any, Dict, List, Optional
-
+import warnings
 from pydantic import BaseModel, BaseSettings, Field
 
+# Comment this function to see Warnings in console
+def warn(*args, **kwargs):
+    pass
+warnings.warn = warn
```
alexbarros, line 10: A feature to disable warnings is desirable, but it should not be enabled/disabled by commenting code, also the default should be having the warnings enabled.

src/pandas_profiling/config.py
```diff
@@ -185,6 +189,22 @@ class Html(BaseModel):
...
+    __float_null = 2
+
+    def fetch_python(self):
+        return self.__default
+
+    def fetch_null_values(self):
+        return self.__num_null
+
+    def fetch_float_values(self):
+        return self.__float_null
```
alexbarros, line 207: I believe Lukas' version is a bit more clear in this part. Can you help me understand why we would want private fields instead of Lukas' version?

src/pandas_profiling/config.py
```diff
@@ -308,6 +333,43 @@ def update(self, updates: dict) -> "Settings":
         update = _merge_dictionaries(self.dict(), updates)
         return self.parse_obj(self.copy(update=update))
 
+class PandasSettings(Settings):
+    pass
+
+class SparkSettings(Settings):
```
alexbarros, line 339: I believe this would make more sense in the spark-branch

tests/issues/test_issue983.py
```diff
@@ -0,0 +1,8 @@
+import pandas as pd
+from pandas_profiling import ProfileReport
+import numpy as np
+df = pd.DataFrame([1, 1, np.nan], columns=["a"])
+
+profile = ProfileReport(df, title="Pandas Profiling Report", minimal=True)
+
+print(profile.to_json())
```
alexbarros, line 8: we should validate the expected behavior for each encoding

alexbarros (changes requested):
Good job, but we need to fix some of the issues with the warnings and the tests, and improve the enum.

fabclmnt:
@Vamp1899 can you please update your commit message as well with the correct and expected format?

PR #1023, opened 2022-08-24
fix: pandas groupby ignoring duplicates with nans

### LINKED ISSUE #1012: Incorrect duplicate rows count
### Current Behaviour

The duplicated rows count is different between pandas and pandas-profiling when there are nan's in columns

### Expected Behaviour

The count should be equal

### Data Description

I attach a simple example

image

```Python
import pandas as pd
import numpy as np

df = pd.DataFrame({"a": [np.nan, np.nan, 2], "b": [1, 1, 3]})
sum(df.duplicated())

profile = ProfileReport(df, title="Pandas Profiling Report")
```

### pandas-profiling version

3.2.0

### Dependencies

```Text
numpy==1.22.4
pandas==1.3.3
```

### OS

_No response_

### Checklist

- [X] There is not yet another bug report for this issue in the [issue tracker](https://github.com/ydataai/pandas-profiling/issues)
- [X] The problem is reproducible from this bug report. [This guide](http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) can help to craft a minimal bug report.
- [X] The issue has not been resolved by the entries listed under [Common Issues](https://pandas-profiling.ydata.ai/docs/master/pages/support_contrib/common_issues.html).

alexbarros:
Fixes #1012 
It is a bug with the pandas groupby (which is used to showcase examples of duplicated rows) that does not consider the missing values.

### DIFF
--- a/src/pandas_profiling/model/pandas/duplicates_pandas.py
+++ b/src/pandas_profiling/model/pandas/duplicates_pandas.py
@@ -35,7 +35,7 @@ def pandas_get_duplicates(
             duplicated_rows = df.duplicated(subset=supported_columns, keep=False)
             duplicated_rows = (
                 df[duplicated_rows]
-                .groupby(supported_columns)
+                .groupby(supported_columns, dropna=False)
                 .size()
                 .reset_index(name=duplicates_key)
             )


fabclmnt (approved):
🚀

PR #1024, opened 2022-08-25
feat: add sortby parameter to timeseries mode

alexbarros:
Add a parameter to sort the data frame for time-series, the parameter is only used if the `tsmode` is active.
```python
profile = ProfileReport(df, tsmode=True, sortby="<column_name>")
```


fabclmnt (approved):
🚀

PR #1030, opened 2022-08-31
feat: multi entity time series heatmap

alexbarros:
Auxiliar function that showcases the distribution of each entity's data across the time axis.

```python
from pandas_profiling.visualisation.plot import timeseries_heatmap
timeseries_heatmap(dataframe, "entity_column", "time_column")
# if the time column is not available the index will be used as the timestamp
```
image

### DIFF
--- a/src/pandas_profiling/profile_report.py
+++ b/src/pandas_profiling/profile_report.py
@@ -139,7 +139,7 @@ class ProfileReport(SerializeReport, ExpectationsReport):
         else:
             return df
 
-    def timeseries_heatmap(
+    def _timeseries_heatmap(
         self,
         dataframe: pd.DataFrame,
         entity_column: str,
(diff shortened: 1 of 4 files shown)


src/pandas_profiling/profile_report.py
```diff
@@ -136,6 +139,77 @@ def __initialize_dataframe(
         else:
             return df
 
+    def _timeseries_heatmap(
```
fabclmnt, line 142: @alexbarros the heatmap function should be an auxiliary function which means shoudn't belog to the ProfileReport class. 

Please add it as a separate function to another python file to be made available independently from the ProfileReport. Let's keep the class method to be related to the Report itself like to_file, etc.

src/pandas_profiling/visualisation/plot.py
```diff
@@ -511,3 +511,27 @@ def plot_acf_pacf(config: Settings, series: pd.Series, figsize: tuple = (15, 5))
     plot_pacf(series.dropna(), lags=lag, ax=axes[1], title="PACF", method="ywm")
 
     return plot_360_n0sc0pe(config)
+
+
+def plot_timeseries_heatmap(
```
fabclmnt, line 516: This is an auxiliary function that does not belong to the ProfileReport, for that reason please consider adding this to be added outside of the methods that are specific to the ProfileReport.

fabclmnt (changes requested):

fabclmnt (approved):

Issue #944, opened 2022-03-24
Alert inconsistency for 'High correlation' in JSON vs HTML Report


Samuele-Campioli:
**Describe the bug**
I don't know if it's feature or a bug, but i think it's the latter. When i use the normal ProfileReport it return the complete Alerts messages, but if i print the alerts in a json format the Alerts result only partials.

**To Reproduce**
```Python
import pandas as pd
import pandas_profiling

file_name = "https://raw.githubusercontent.com/uiuc-cse/data-fa14/gh-pages/data/iris.csv"
df = pd.read_csv(file_name)
ProfileReport(df)

ProfileReport(df).to_file('./test_file.json')
```
Result from the first command:
<html>
<body>
<p class="h4" style="box-sizing: border-box; stroke-linecap: butt; stroke-linejoin: round; margin: 10px 0px; font-family: &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-weight: 500; line-height: 1.1; color: rgb(51, 51, 51); font-size: 18px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Alerts</p>

Dataset has 2 (1.3%) duplicate rows | Duplicates
-- | --
sepal_length is highly correlated with petal_length and 1 other fields | High correlation
petal_length is highly correlated with sepal_length and 1 other fields | High correlation
petal_width is highly correlated with sepal_length and 1 other fields | High correlation
sepal_length is highly correlated with petal_length and 1 other fields | High correlation
petal_length is highly correlated with sepal_length and 1 other fields | High correlation
petal_width is highly correlated with sepal_length and 1 other fields | High correlation
sepal_length is highly correlated with petal_length and 1 other fields | High correlation
petal_length is highly correlated with sepal_length and 1 other fields | High correlation
petal_width is highly correlated with sepal_length and 1 other fields | High correlation
sepal_length is highly correlated with sepal_width and 3 other fields | High correlation
sepal_width is highly correlated with sepal_length and 3 other fields | High correlation
petal_length is highly correlated with sepal_length and 3 other fields | High correlation
[...]

Issue #647, opened 2020-12-22
Value Counts Bin


vinutshetty:
Hi,

Is there a setting in config file to restrict "value counts" to say 50 max. I profiled 10M records and it created 3GB big JSON file. Mainly because of "value counts". 

Is there a way to restrict say top 10 value counts.

Issue #637, opened 2020-12-01
Document fields in the JSON report


dclong:
**Missing functionality**

The JSON report has lots of fields but unfortunately, they are very confusing and hard to figure out their meanings. For example, I thought `is_unique` tells whether a column has unique value or not, but it doesn't seem to be the case. Also, I have no idea what `n_unique` means. Based on the name, I'd guess it is then number of unique values, but it doesn't seem to be the case either. 

**Proposed feature**

Document fields in the JSON report. 

**Alternatives considered**

**Additional context**

sbrugman:
Agreed, it would be a great addition to have a JSON schema. Contributions welcome.

To answer your question: `n_unique` is the number of unique values (i.e. occur exactly once). You might be looking for `n_distinct`, which is the number of distinct values.

dclong:
Thank you, @sbrugman! I confused myself on `unique` vs `distinct`.

Issue #636, opened 2020-12-01
Support customized schema when output a report to a JSON file


dclong:
**Missing functionality**

In the case of profiling a large data frame (many rows and columns),
a large (a few GBs) JSON report can be generated,
which is not convenient for further filtering based on the JSON report. 
It would be great if there's a way to control the output schema off a JSON report,
so that some fields containing large data can be skipped to reduce the size of the JSON report. 
There is no way to control the output schema of a JSON report. 

**Proposed feature**

Support customizing the output schema. Allowing skipping some fields. 

**Alternatives considered**

**Additional context**

sbrugman:
Agreed, good suggestion. Could be a follow-up from #637

PR #1069, opened 2022-09-25
Report comparisons

sbrugman:
This pull request introduces the report comparison functionality into ``pandas-profiling``. Various other fixes/changes are included.

- [x] Rebase on develop
- [x] Add feature tests
- [x] Ensure tests are passing
- [x] User testing of reports created in CI/CD
- [ ] Refactor into consistent data structure

---
dataset-comparison

---

``pandas-profiling`` can be used to compare multiple version of the same dataset.
This is useful when comparing data from multiple time periods, such as two years.
Another common scenario is to view the dataset profile for training, validation and test sets in machine learning.

The following syntax can be used to compare two datasets:

```python

train_df = pd.read_csv("train.csv")
train_report = ProfileReport(train_df, title="Train")

test_df = pd.read_csv("test.csv")
test_report = ProfileReport(test_df, title="Test")

comparison_report = train_report.compare(test_report)
comparison_report.to_file("comparison.html")
```

---

`Settings.html.style.primary_color` is replaced by a sequence of colors `Settings.html.style.primary_colors`. For backwards compatibility, the first element of the sequence is still accessible through the `primary_color` attribute, however deprecation in the future is desired. The comparison feature requires to change the configuration of colors. Where previously there was no distinction between the parts specific to the report, and parts specific to a single dataset summary, this distinction is introduced now. It could be misleading to simply use the first of the colors as a report color (The first `#377eb8` now is close to the default report color `#337ab7`). Future work could go into configuring the report color and chosing better defaults.

---

Refactoring the summary datastructure is partially out of scope for this PR. (For this there are other design decisions that have been considered before), see: https://github.com/ydataai/pandas-profiling/issues/1102


src/pandas_profiling/report/formatters.py
```diff
@@ -3,12 +3,23 @@
...
 import re
 from datetime import timedelta
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import numpy as np
 from markupsafe import escape
 
 
+def list_args(func: Callable) -> Callable:
```
akx, line 12: This could definitely use a docstring... By the looks of it, it... decorates the function `f(a, ...)` to return `map(f(...), a)` if argument `a` is a list..? Sounds weird.

src/pandas_profiling/report/structure/correlations.py
```diff
@@ -68,29 +68,65 @@ def get_correlation_items(config: Settings, summary: dict) -> Optional[Renderabl
...
+                tbl = Container(
+                    diagrams + [desc],
+                    anchor_id=key,
+                    name=name,
+                    sequence_type="batch_grid",
+                    batch_size=len(config.html.style._labels) + 1,
+                )
+                cont.append(tbl)
+            else:
+                raise RuntimeError("Correlation descriptions missing.")
```
akx, line 102: It would probably be nicer to make this an "early-out" check: `if not description: raise...` instead of nest it as an `else` of a long `if`.

src/pandas_profiling/report/structure/overview.py
```diff
@@ -189,15 +213,48 @@ def get_dataset_column_definitions(definitions: dict) -> Container:
...
+    Returns:
+        Alerts renderable object
+    """
+    # add up alerts from multiple reports
+    if isinstance(alerts, tuple):
+        count = 0
+        combined_alerts = defaultdict(dict)
+        for i, a in enumerate(alerts):
+            for alert in a:
+                combined_alerts[f"{alert.alert_type}_{alert.column_name}"][i] = alert
```
akx, line 232: This seems to make the keys of `combined_alerts` keys "ragged"?

src/pandas_profiling/report/structure/overview.py
```diff
@@ -189,15 +213,48 @@ def get_dataset_column_definitions(definitions: dict) -> Container:
...
+            count += len(
+                [alert for alert in a if alert.alert_type != AlertType.REJECTED]
+            )
+
+        return Alerts(
+            alerts=combined_alerts,
```
akx, line 238: Hmm – this is passing a `defaultdict` as `alerts`, which is annotated to be `List[Alert]`.

src/pandas_profiling/report/structure/report.py
```diff
@@ -67,25 +85,47 @@ def render_variables_section(config: Settings, dataframe_summary: dict) -> list:
...
+            }
+        else:
+            alerts = tuple(
+                [alert.fmt() for alert in abc if alert.column_name == idx]
+                for abc in dataframe_summary["alerts"]
+            )  # type: ignore
+
+            alert_fields = {
+                field
+                for abc in dataframe_summary["alerts"]
```
akx, line 117: `abc` doesn't sound like the greatest of variable names...

src/pandas_profiling/report/structure/report.py
```diff
@@ -276,23 +402,26 @@ def get_report_structure(config: Settings, summary: dict) -> Root:
...
+        duplicate_items = get_duplicates_items(config, summary["duplicates"])
         if len(duplicate_items) > 0:
             section_items.append(
                 Container(
                     items=duplicate_items,
-                    sequence_type="list",
+                    sequence_type="batch_grid",
+                    batch_size=len(duplicate_items)
+                    if isinstance(duplicate_items, list)
+                    else 1,
```
akx, line 424: Smells like this should be parenthesized for ease of reading (but really it would be better if `duplicate_items` just always was a list)

src/pandas_profiling/visualisation/plot.py
```diff
@@ -141,7 +180,7 @@ def get_correlation_font_size(n_labels: int) -> Optional[int]:
     """Dynamic label font sizes in correlation plots
 
     Args:
-        n_labels: the number of labels
+        n_labels: the number of _labels
```
akx, line 183: I don't think `_labels` should be underscore-prefixed here.

src/pandas_profiling/visualisation/plot.py
```diff
@@ -223,7 +262,7 @@ def scatter_complex(config: Settings, series: pd.Series) -> str:
     plt.ylabel("Imaginary")
     plt.xlabel("Real")
 
-    color = config.html.style.primary_color
+    color = config.html.style.primary_colors[0]
```
akx, line 265: This doesn't seem to be required since the `Style` class has the `primary_color` property anyway..?

akx (commented):
* :raised_hands: for small granular commits :)

* There are a bunch of long `if isinstance(..., tuple)` / ... style branches which seem to partially duplicate code and as such will eventually cause drift between the branches. I suppose these are to adapt rendering for one report or more than one report when comparing.
  * Similarly there are functions like `_get_n` for that, not to mention the `@list_args` decorator...
  * This also repeats for the templates, with `if alerts is mapping` sort of branches that also have a slightly different formatting..?
  * It would probably be nicer to not have those branches, but adapt the code everywhere to the general case of 1 or more reports.
  * a7b4c114b61bc4a7e5b2ca02625c512d29d09a50 wouldn't probably be required separately if that was the case.

* Should there be a test for adfa846fc122dcac247593b5ac3bf703a2e09bc1? 
* Should the (mechanical?) reindentation of the template(s) happen in a separate commit?
* There are some commits that don't really look `refactor:` to me, such as cbfc01828f689abb2402b309a9cbe6149b24013a, 85ad2c5f7e05c9ff3bbb0b355029d4baf869035c, d7c583cc3bce75b4d16e232306650b2099c7da98, 3c97aed08dca8916f26360199c974fb06fb38c4d, 4828c15124e809cc0f21b34435f32967560b5de0, 6354e906adad42dc205c374c91a39f332b97be51, fe181eaacb746c976f58fdce0bd8021f4a6bf081 (what does "namespace invariant type check" actually mean?)
* 54d24ea4c8acf8cb48fc52e6a48ba54b9c9418ca's commit message doesn't make much sense to me..? Should the new? behavior be documented?
* To paraphrase @fabclmnt, can you follow the correct PR naming pattern?

src/pandas_profiling/model/pandas/table_pandas.py
```diff
@@ -20,7 +20,7 @@ def pandas_get_table_stats(
     Returns:
         A dictionary that contains the table statistics.
     """
-    n = len(df)
+    n = len(df) if not df.empty else 0
```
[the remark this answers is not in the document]
sbrugman: `len(df)` is non-zero when the `df.empty` is True, but the index is not. See `test_dataframe_empty.py`

src/pandas_profiling/visualisation/plot.py
sbrugman: We should deprecate `primary_color`. Not intending to make this feature breaking backwards compatibility. Changes in preparation of removal.

src/pandas_profiling/report/structure/report.py
sbrugman: This check was outdated, `duplicate_items` always is a list

src/pandas_profiling/profile_report.py
```diff
@@ -256,8 +286,6 @@ def get_rejected_variables(self) -> set:
     def to_file(self, output_file: Union[str, Path], silent: bool = True) -> None:
         """Write the report to a file.
 
-        By default a name is generated.
-
```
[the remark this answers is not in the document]
sbrugman: The docstring was outdated, the functionality was removed longer ago.

src/pandas_profiling/report/formatters.py
sbrugman: This was added for backwards compatibility. It simply extends the function to support providing the first argument as list. Ideally the data structure is consistent for a single and multiple reports, this is out of scope for this PR (it is sizable already).

src/pandas_profiling/report/structure/correlations.py
sbrugman: The condition turned out to be not needed at all, descriptions are always present (and should be).

src/pandas_profiling/config_default.yaml
```diff
@@ -206,6 +206,9 @@ html:
     style:
       theme: null
       logo: ""
-      primary_color: "#337ab7"
+      primary_colors:
+      - "#377eb8"
+      - "#e41a1c"
+      - "#4daf4a"
```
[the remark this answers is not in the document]
sbrugman: Good point, added a note to the PR.

sbrugman:
> * There are a bunch of long `if isinstance(..., tuple)` / ... style branches which seem to partially duplicate code and as such will eventually cause drift between the branches. I suppose these are to adapt rendering for one report or more than one report when comparing.
>   
>   * Similarly there are functions like `_get_n` for that, not to mention the `@list_args` decorator...
>   * This also repeats for the templates, with `if alerts is mapping` sort of branches that also have a slightly different formatting..?
>   * It would probably be nicer to not have those branches, but adapt the code everywhere to the general case of 1 or more reports.
>   * [a7b4c11](https://github.com/ydataai/pandas-profiling/commit/a7b4c114b61bc4a7e5b2ca02625c512d29d09a50) wouldn't probably be required separately if that was the case.

Agreed - added a note to the PR. This is out of scope for the feature itself and requires a redesign of the data structure used for the dataset summaries. 

> * Should there be a test for [adfa846](https://github.com/ydataai/pandas-profiling/commit/adfa846fc122dcac247593b5ac3bf703a2e09bc1)?

Added a test case for this. One of the items on the wishlist is rewriting the test suits entirely.

Regarding the other comments: the commit history is not rewritten yet, this was pending the review of the code.

src/pandas_profiling/compare_reports.py
```diff
@@ -0,0 +1,144 @@
...
+
+    if len(reports) > 2:
+        warnings.warn(
+            "Comparison of more than two reports is in beta. "
+            "Reports will be produced, but may yield unexpected formatting."
+        )
+
+    if any(not isinstance(report, dict) for report in reports):
+        # Use titles as labels
+        labels = [report.config.title for report in reports]  # type: ignore
```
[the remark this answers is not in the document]
sbrugman: The `type: ignore`s are partially needed because mypy does not pick up on the type narrowing. In 3.10 there is the explicit `TypeGuard`. Having the explicit schema for dataset descriptions will also help (as part of the refactor mentioned a couple of times here now). Refactored part of the code to reduce the 'type: ignore's.

src/pandas_profiling/report/structure/overview.py
sbrugman: This could indeed be solved a bit more elegantly (it is the same case with placeholders in the compare file). Refactored the code to be more clear.

src/pandas_profiling/report/formatters.py
akx: While there currently don't seem to be formatters that could "natively" work on lists, they'd need a special case with this decorator, too... 😕

akx:
> Agreed - added a note to the PR. This is out of scope for the feature itself and requires a redesign of the data structure used for the dataset summaries.

My two Euro cents (or more accounting for inflation) I honestly think it would be better to do the work to make the data structures consistent now (or even in a separate PR this could then build upon), as opposed to adding special casing and "hacks" now and cleaning it up later.

sbrugman:
@akx Agreed, the work needs to be done anyway. I've looked for a sensible split in the refactoring. 

To maintain backward compatibility (for now) and not blowing up this PR, the data structure will wrap each 'leaf' in the data structure 'tree' with a list. With that data structure, only one of the two branches (list/no list) are required, and most duplicate code can be removed. To not break backward compatibility for users that rely on the `report.description_set`, we'll only convert the structure once the report is generated.

A follow-up PR will take care of having a well-defined JSON schema fixing https://github.com/ydataai/pandas-profiling/issues/983,  https://github.com/ydataai/pandas-profiling/issues/944, https://github.com/ydataai/pandas-profiling/issues/647,  https://github.com/ydataai/pandas-profiling/issues/637, https://github.com/ydataai/pandas-profiling/issues/636,  https://github.com/ydataai/pandas-profiling/issues/624 and https://github.com/ydataai/pandas-profiling/issues/605.

aquemy (approved):

sbrugman:
@aquemy That's a good point. Actually, in the first designs the delta was a central part. The challenge is providing a meaningful way to present the delta, without discarding information on the absolute compared values. Juxtapositioning is a sensible default: all values are there, the user can easily see the comparison.

In case you have concrete improvements on delta's in mind that increase clarity and are consistent with the layout then they are of course welcome. However this may be hard to know upfront, so we might discover them from users of the basic version.

fabclmnt:
> @aquemy That's a good point. Actually, in the first designs, the delta was a central part. The challenge is providing a meaningful way to present the delta, without discarding information on the absolute compared values. Juxtapositioning is a sensible default: all values are there, and the user can easily see the comparison.
> 
> In case you have concrete improvements on the delta in mind that increase clarity and are consistent with the layout then they are of course welcome. However this may be hard to know upfront, so we might discover them from user research that is working with the basic version.

Although I do find this is relevant and important for the comparison report, I would prefer to keep the first iteration simpler - without the computed delta. My suggestion is to keep this in a separate `feature request`.

I see no changes needed. This version is already looking great! One small request, and as @akx mentioned, please update the PR message to `feat`: (description)

docsrc/source/pages/use_cases/comparing_datasets.rst
```diff
@@ -0,0 +1,52 @@
+==================
+Dataset Comparison
+==================
+
+*This feature was introduced in pandas-profiling 3.4.*
```
fabclmnt, line 5: Please remove this reference from the documentation. 

We can add it later with a changelog reference for instance.

docsrc/source/pages/use_cases/comparing_datasets.rst
```diff
@@ -0,0 +1,52 @@
...
+
+    comparison_report = compare([train_report, validation_report, test_report])
+
+    # Obtain merged statistics
+    statistics = comparison_report.get_description()
+
+    # Save report to file
+    comparison_report.to_file("comparison.html")
+
+Note that generating reports for three or more datasets is not (yet) fully supported.
```
fabclmnt, line 45: Can you update this sentence to the following:

> Note that this functionality only ensures the support to generate reports to compare two datasets.

docsrc/source/pages/use_cases/comparing_datasets.rst
```diff
@@ -0,0 +1,52 @@
...
+It is possible to obtain the statistics - the report may have formatting issues.
+One of the settings that can be changed to improve the formatting is ``settings.report.precision``.
+As a rule of thumb the value 10 can be used for a single report, 8 for comparing two reports and 5 for comparing three reports.
+
```
fabclmnt, line 49: _I suggest changing the above paragraph to the following:_

In case your report has formatting issues, you can improve it by tweaking the settings. One of the settings that can be changed is ``settings.report.precision``. As a rule of thumb, the value 10 can be used for a single report and 8 for comparing two reports.

docsrc/source/pages/use_cases/comparing_datasets.rst
```diff
@@ -0,0 +1,52 @@
...
+.. pull-quote::
+
+    ⌛ Interested in uncovering more temporal patterns? Check out `popmon <https://github.com/ing-bank/popmon>`_.
```
fabclmnt, line 52: Can you please remove the reference to popmon from the comparison report documentation?

fabclmnt (dismissed):
Small documentation changes request.

aquemy (approved):

PR #1070, opened 2022-09-27
fix: make tangled-up-in-unicode an optional dependency

akx:
Knowing the exact Unicode properties is likely not a priority for all users. Furthermore, unicodedata as of Python 3.10 is based on UCD 13, so the category data in there ought to be fairly recent.

This should probably be documented (as well as the `[unicode]` extra), but I'm not sure what the best place for it would be.

Refs https://github.com/dylan-profiler/tangled-up-in-unicode/issues/10 – importing `tangled` will create 2 gigabytes of 
bytecode files and takes a while.

### DIFF
--- a/src/pandas_profiling/model/pandas/describe_categorical_pandas.py
+++ b/src/pandas_profiling/model/pandas/describe_categorical_pandas.py
@@ -53,7 +53,14 @@ def counter_to_series(counter: Counter) -> pd.Series:
 
 
 def unicode_summary_vc(vc: pd.Series) -> dict:
-    from tangled_up_in_unicode import block, block_abbr, category, category_long, script
+    try:
+        from tangled_up_in_unicode import block, block_abbr, category, category_long, script
+    except ImportError:
+        from unicodedata import category
+        block = lambda char: "(unknown)"
+        block_abbr = lambda char: "(unknown)"
+        category_long = lambda char: "(unknown)"
+        script = lambda char: "(unknown)"
 
     # Unicode Character Summaries (category and script name)
     character_counts = get_character_counts_vc(vc)
--- a/README.md
+++ b/README.md
@@ -147,8 +147,22 @@ Additional details, including information about widget support, are available [o
 You can install using the `pip` package manager by running:
 
 ```sh
-pip install -U pandas-profiling[notebook]
+pip install -U pandas-profiling
 ```
+
+#### Extras
+
+The package declares "extras", sets of additional dependencies.
+
+* `[notebook]`: support for rendering the report in Jupyter notebook widgets.
+* `[unicode]`: support for more detailed Unicode analysis, at the expense of additional disk space.
+
+Install these with e.g.
+
+```sh
+pip install -U pandas-profiling[notebook,unicode]
+```
+
 
 ### Using conda
 [![Conda Downloads](https://img.shields.io/conda/dn/conda-forge/pandas-profiling.svg)](https://anaconda.org/conda-forge/pandas-profiling)
--- a/docsrc/source/pages/getting_started/installation.rst
+++ b/docsrc/source/pages/getting_started/installation.rst
@@ -21,8 +21,7 @@ You can install using the ``pip`` package manager by running:
 
 .. code-block:: console
 
-    pip install -U pandas-profiling[notebook]
-    jupyter nbextension enable --py widgetsnbextension
+    pip install -U pandas-profiling
 
 If you are in a notebook (locally, LambdaLabs, Google Colab or Kaggle), you can run:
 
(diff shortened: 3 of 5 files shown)


fabclmnt:
Hi @akx,

although I understand where the requirement comes from, not always the user will no ahead of they won't need to install the package. This would be great under the assumption that the user will already now the dataset has text to be analyzed when setting an environment with pandas-profiling. 

Furthermore, the fact that **visions** is a requirement of **pandas-profiling** and  **tangled-up-in-unicode** is a vision's dependecy, this wouldn't prevent the problem you mention. 

Instead of making it optional, I would suggest adding to the roadmap pursuing a different option (maybe another package, etc.)

I'll close this PR. Can I invite you to explore other options to solve your request?

akx:
Hi @fabclmnt, thanks for getting back to me!

1. If we clearly document that more in-depth character analysis is possible if one installs `pandas-profiling[unicode]` (at the expense of significant disk space), that should take care of that, I think?
2. `visions` hasn't actually _required_ tangled since May 2021, the dependency is there for no reason. I opened the PR https://github.com/dylan-profiler/visions/pull/197 in `visions` prior to opening this just for that. 😄

akx:
So @fabclmnt - with the above provisions, could this be reopened (for discussion?)?

fabclmnt:
@akx the 2GB is a pain indeed and as I've mentioned before, we are happy to consider any other suggestion to have it sorted, different from the one suggested by this PR. 

This suggestion does not fit the profile of the user and use-cases for which this package was designed for.

fabclmnt:
> > This suggestion does not fit the profile of the user and use-cases for which this package was designed for.
> 
> @fabclmnt I'm not sure I understand – can you explain what the profile and the use-cases are? For users who require precise knowledge of the distribution of Unicode blocks, categories and scripts in their data, they can install `pandas-profiling[unicode]`. For the rest, everything remains the same.

The point is precisely what you've mentioned, the user would have to know ahead that he needs or not to use the unicode at the moment of the installation which is not the case for our audience. I have particular examples and request of users that for the same use case they explore different datasets (with the need for the precise knowledge of the distribution of unicode blocks) but they only know it, whenever they use PP for the initial exploration and data understanding.

fabclmnt:
Re-openning this PR - after having a look, this is indeed the best option. 

Can you please add a note to the documentation regarding this change?

fabclmnt (approved):

fabclmnt (changes requested):
I'll just ask you to change the commit message and change to the `develop` branch.

akx:
@fabclmnt 👍 Changed base to `develop`, rebased, and amended readme and docs too :)

src/pandas_profiling/model/pandas/describe_categorical_pandas.py
```diff
@@ -53,7 +53,14 @@ def counter_to_series(counter: Counter) -> pd.Series:
 
 
 def unicode_summary_vc(vc: pd.Series) -> dict:
-    from tangled_up_in_unicode import block, block_abbr, category, category_long, script
+    try:
+        from tangled_up_in_unicode import block, block_abbr, category, category_long, script
```
sbrugman, line 57: This check should be earlier (at the condition `if config.vars.cat.characters`). The report renderer is already able to deal with empty unicode sets. 

If unicode is available, but the library is not installed, then a warning should be given.

sbrugman:
Apart from the location of the check, lgtm.

src/pandas_profiling/model/pandas/describe_categorical_pandas.py
akx: @sbrugman Sorry, what do you mean with "if unicode is available"? 

Do you mean that if `config.vars.cat.characters` is enabled and tangled isn't, we should raise a warning and not do any of the character checks? Since `unicodedata` (built-in in Python) provides `category`, we can always provide category data if nothing else.

fabclmnt (approved):
🚀 looking good to go.

src/pandas_profiling/model/pandas/describe_categorical_pandas.py
fabclmnt: @akx I do agree with you. I don't think we should raise a warning in this situation given that we have built-in Python `unicodedata`.

aquemy:
Given the fact that we need to release and that the concern raised by @sbrugman is not a blocker or vital, I merge this.

Kudos to @akx !

src/pandas_profiling/model/pandas/describe_categorical_pandas.py
sbrugman: For future reference: the builtin `unicodedata`'s category returns a category code. The `category_long` in `tangled-up-in-unicode` looks up the more descriptive longer version useful for users. Might be worth adding a link/hint when some unicode properties are uncomputed.

PR #1074, opened 2022-09-28
chore(actions): update github actions flow

vascoalramos:
This pull request introduces an update to the overal GitHub and GitHub actions flow, from the uniformization with the standard YData processes.

The overall process follows the flow expressed in the image below:
git-workflow-diagram

Regarding CI/CD:
- When there is a Pull Request:
    - Lint commit message and source code
    - Execute multi-software tests
- When there is a merge to dev:
    - Create a Prerelease tagged with `major.minor.patch.pre#number"
    - Execute multi-software tests
- When there is a merge to master:
    - Create a Prerelease tagged with `major.minor.patch.rc#number"
    - Create a Draft Release
    - Execute multi-software tests
- When there is a Prerelease:
    - Upload assets to Release
- When there is a Release:
    - Upload assets to Release
    - Publish packages to public PyPi


src/pandas_profiling/version.py
```diff
@@ -1,2 +1,2 @@
 """This file is auto-generated by setup.py, please do not alter."""
-__version__ = "3.3.0"
+__version__ = "3.3.1"
```
alexbarros, line 2: > """This file is auto-generated by setup.py, please do not alter."""

.pre-commit-config.yaml
```diff
@@ -9,9 +12,9 @@ repos:
     hooks:
     - id: nbqa-black
     - id: nbqa-isort
-      args: [ --nbqa-mutate, --profile=black, --project=pandas_profiling ]
+      args: [ --profile=black, --project=pandas_profiling ]
```
alexbarros, line 15: why `nbqa-mutate` was dropped?

.github/workflows/triage.yml
```diff
@@ -1,18 +1,26 @@
...
-        with:
-          add-labels: "needs-triage"
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
-          ignore-if-labeled: true
+    - name: Label issues
+      uses: andymckay/labeler@master
+      with:
+        add-labels: "needs-triage"
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        ignore-if-labeled: true
```
alexbarros, line 26: is there any reason for these style changes?

.github/workflows/tests.yml
```diff
@@ -39,42 +50,50 @@ jobs:
...
+    - run: |
+        pip install --upgrade pip setuptools wheel
+        pip install -r requirements.txt "${{ matrix.pandas }}" "${{ matrix.numpy }}"
+        pip install -r requirements-test.txt
+
+    - run: make install
+
+    - run: make test
+
+
```
alexbarros, line 94: this does not seem to change anything, is it just style changes? If yes, why?

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,104 @@
...
+    - uses: actions/download-artifact@v2
+      with:
+        name: built-artifacts
+        path: dist/
+
+    - uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        user: ${{ secrets.PRIVATE_PYPI_USERNAME }}
+        password: ${{ secrets.PRIVATE_PYPI_PASSWORD }}
+        repository_url: ${{ secrets.PRIVATE_PYPI_URL }}
```
alexbarros, line 104: do we need to publish pre-releases to pypi?

src/pandas_profiling/version.py
vascoalramos: I didn't have a recollection of editing this file, so I checked it. The change was done by the CI/CD code formatting step
https://github.com/ydataai/pandas-profiling/pull/1074/commits/91d4d5739e984412ab503c8003bd2b5818222d23

.pre-commit-config.yaml
vascoalramos: while testing the new approach, I noted a warning during execution stating this optional flag has been dropped since it was now the default behavior. I can revert this change if you fell more comfortable

.github/workflows/triage.yml
vascoalramos: No, my mistake. Fixed it

.github/workflows/tests.yml
vascoalramos: It was a formatting mistake, fixed it.

.github/workflows/release.yml
```diff
@@ -25,113 +110,55 @@ jobs:
...
-      runs-on: ubuntu-latest
-      needs: lint
-      steps:
-      - uses: actions/checkout@v2
-
-      - name: Configurating Git
-        run: |
-          git config --local user.email "action@github.com"
-          git config --local user.name "GitHub Action"
-          git config --local core.autocrlf false
```
portellaa, line 71: shouldn't we leave this? 🤔

.github/workflows/tests.yml
```diff
@@ -1,6 +1,17 @@
...
+on:
+  pull_request:
+    branches:
+    - master
+  push:
+    branches: 
+    - master
+    - develop
```
portellaa, line 11: this will only run tests in the master and developer branches and in pull requests that goes to master, right? what about a pull request to develop?

honestly i think we should test it in the pull request, no matter the destination, that is the point that matters the most, in my opinion, if something fails you are still in time to fix it and most of all it blocks the PR of being merged (it should btw, we probably have to update the settings in the github)

src/pandas_profiling/version.py
portellaa: this should have been like this in the master, when the release was done, which means that the latest release is wrong 😕 https://github.com/ydataai/pandas-profiling/blob/master/src/pandas_profiling/version.py

it should be 3.3.1

portellaa: never mind, my mistake. why did it change? do you know?

vascoalramos: it was due to the setup.py have already the 3.3.1 as a version. the automatic update of the version.py was moved to the PR CI/CD since all changes should happen as PR, even version bumps

alexbarros (approved):

.github/workflows/prerelease.yml
fabclmnt: Yup I don't see any issue with having the pre-release published.

.releaserc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "tagFormat": "${version}",
```
portellaa, line 2: please remove this format, this will break the cycle.

the current version format is `vX.Y.Z`, if you remove the v it will start from the beginning.

portellaa (changes requested):

.releaserc.json
vascoalramos: will do!

portellaa (approved):
looks good. nice job 🎉

PR #1076, opened 2022-09-29
feat: add support for Pandas 1.5

### LINKED ISSUE #1075: Feature Request: support of pandas 1.5
### Missing functionality

pandas-profiling 3.3.0 requires pandas version lower than 1.5.

### Proposed feature

Compatibility with pandas 1.5

### Alternatives considered

_No response_

### Additional context

_No response_

akx:
Some imports had to be fixed up.

* `make test` passes locally with flying colors with Pandas 1.5.

Fixes #1075

Following @sbrugman's review, I also added a new `pandas_version_info()` function, and at the same time got rid of outdated `pandas.__version__` checks (since this library requires Pandas 1.2 since 3524c912ad16e2bdc7ce6225fdc32360b2335cba).


src/pandas_profiling/model/pandas/correlations_pandas.py
```diff
@@ -73,12 +73,15 @@ def pandas_cramers_compute(
 ) -> Optional[pd.DataFrame]:
     threshold = config.categorical_maximum_correlation_distinct
 
-    categoricals = {
-        key
-        for key, value in summary.items()
-        if value["type"] in {"Categorical", "Boolean"}
-        and value["n_distinct"] <= threshold
-    }
+    # Index must not be a set, so sort it into a list
```
fabclmnt, line 76: Can you please explain why the sorting is required with pandas 1.5 support?

fabclmnt (changes requested):
@akx please change the merge for develop branch instead of master. 

Also can I ask you to follow the Pr naming pattern. 
`feat: add support for Pandas 1.5`

src/pandas_profiling/model/pandas/correlations_pandas.py
akx: Sure:

* https://github.com/pandas-dev/pandas/commit/cde13db86a2
* https://github.com/pandas-dev/pandas/pull/47231
* https://github.com/pandas-dev/pandas/issues/47215

akx:
@fabclmnt Done. Maybe the GitHub default branch should be switched to `develop`..?

src/pandas_profiling/model/pandas/correlations_pandas.py
fabclmnt: So the sorting is not related to the support for pandas 1.5. Just a nice to have based on my understanding.

fabclmnt:
@akx the branch was set to default for `master` for purposes of examples and tutorial checking in GitHub. Otherwise, when you land in GitHub you'll be facing a branch with examples that are not yet available through PyPI. Unfortunately, it was a common misconception.

I understand it might not be ideal, but it was the choice for now. Thank you for understanding!

src/pandas_profiling/model/pandas/correlations_pandas.py
akx: It is related to 1.5 - the PR linked above that prevents using sets for indexes only landed in 1.5.

It is possible the order changes, but then again that used to be the case prior to this change, since sets are inherently unordered.

If you like, I can change `sorted` to `list` instead, to preserve the possibly arbitrary order the set is iterated in.

fabclmnt: Just wanted to double-check whether this was something new, or a match of former functionality. 

Thank you for the clarification! Yes, that was my understanding and I would prefer to preserve for now the inherited arbitrary order.

fabclmnt (approved):
🚀

src/pandas_profiling/model/pandas/correlations_pandas.py
akx: Updated to use a plain `list()`, with an explanatory comment. 👍

src/pandas_profiling/model/correlations.py
```diff
@@ -5,7 +5,12 @@
 import numpy as np
 import pandas as pd
 from multimethod import multimethod
-from pandas.core.base import DataError
+
+try:
```
sbrugman, line 9: Instead of try/except, check using pd.__version__ (which is already available)
This prevents unnecessary work for older pandas versions.

tests/unit/test_utils.py
```diff
@@ -37,10 +37,13 @@ def test_read_pandas_csv():
 
 def test_read_pandas_json():
     p = Path("dataframe.json")
-    with pytest.raises(ValueError) as e:
+    with pytest.raises((ValueError, FileNotFoundError)) as e:
```
sbrugman, line 40: Split into two tests depending on the pandas version (one could be dropped eventually)

src/pandas_profiling/model/pandas/describe_numeric_pandas.py
```diff
@@ -2,7 +2,12 @@
 
 import numpy as np
 import pandas as pd
-from pandas.core.arrays.integer import _IntegerDtype
+
+try:
```
sbrugman, line 6: Same as earlier

sbrugman (changes requested):
Thanks for taking this on, couple of small changes.

src/pandas_profiling/model/correlations.py
akx: `pandas.__version__` is not being used for any other check in `pandas-profiling` so far. I'm not sure it would be wise to add it here.

With "unnecessary work", do you mean a single `except:` clause processing? That would be done exactly one over the runtime of a program, and takes a negligible time nevertheless...

tests/unit/test_utils.py
akx: This would again require a version check (which haven't been used in this library so far)... I don't think it's worth the added test complexity.

src/pandas_profiling/model/pandas/describe_numeric_pandas.py
akx: Same reply as earlier :)

src/pandas_profiling/model/correlations.py
sbrugman: Not in the library - there was no need - but it is used in the tests, and in the type system (visions), e.g.:
```def pandas_version():
    return tuple(int(s) for s in pd.__version__.split("."))


@pytest.mark.skipif(
    pandas_version() <= (1, 1, 0), reason="requires pandas 1.1.1 or higher"
)
```

You're right that in this case performance would probably not matter much.  I prefer this syntax as it's more explict

akx:
@sbrugman Changes enacted. Can't re-request review mechanically from some reason, so here's a comment to re-request review 😁

sbrugman (approved):
Great work, thank you!

PR #1083, opened 2022-09-30
feat: association correlation

jtook:


src/pandas_profiling/version.py
```diff
@@ -1,2 +1,2 @@
 """This file is auto-generated by setup.py, please do not alter."""
-__version__ = "3.3.0"
+__version__ = "3.3.1"
```
fabclmnt, line 2: There's something wrong with the PR, you are "dragging" files not expected. 

@JamesTook can you please check if you've rebased correctly?

akx: FWIW, this change happened to me in PR #1076 too, but I just decided not to stage it. I'm not sure which automagical process alters this...

src/pandas_profiling/report/structure/correlations.py
```diff
@@ -55,12 +55,19 @@ def get_correlation_items(config: Settings, summary: dict) -> Optional[Renderabl
     The empirical estimators used for Cramér's V have been proved to be biased, even for large samples.
     We use a bias-corrected measure that has been proposed by Bergsma in 2013 that can be found <a href='http://stats.lse.ac.uk/bergsma/pdf/cramerV3.pdf'>here</a>."""
 
+    association_description = """The association metric is an easily interpretable pairwise column metric of the following mapping:
```
sbrugman, line 58: A link to reference material would be nice here

docsrc/source/pages/advanced_usage/available_settings.rst
```diff
@@ -96,6 +96,7 @@ For instance, to disable all correlation computations (may be relevant for large
             "kendall": {"calculate": False},
             "phi_k": {"calculate": False},
             "cramers": {"calculate": False},
+            "association": {"calculate": False},
```
sbrugman, line 99: The metric should also be added to `../tables/config_correlations.csv`

src/pandas_profiling/model/pandas/correlations_pandas.py
```diff
@@ -100,6 +113,49 @@ def pandas_cramers_compute(
...
+        key for key, value in summary.items() if value["type"] == "Numeric"
+    ]
+    categorical_columns = [
+        key
+        for key, value in summary.items()
+        if value["type"] in {"Categorical", "Boolean"}
+        and value["n_distinct"] <= threshold
+    ]
+    df_discretized = Discretizer(
+        DiscretizationType.UNIFORM, n_bins=10
```
sbrugman, line 131: The number of bins could be a configuration parameter above

sbrugman (changes requested):
See comments for requested changes.

In addition, please squash the commits.

fabclmnt (changes requested):
@jtook please have the linter warnings solved. Also please change the merge branch to `develop` and not `master`.

aquemy (changes requested):
Hi,

The approach taken here is not to replace the other metrics that are still available via the `correlation_name` parameter. Instead, we are adding a more unified view that uses Pearson and Cramer's V in the same matrix.
Therefore, I suggest to rename 'association' by 'unified'. But because this is not a 'raw' metric, I would strongly encourage you to comment wherever it is needed to fully explain what it does in the doctring of `calculate_correlation` below the list of metrics .


Few general comments:

1. Please, check the files that you stage. Do not use `git commit -a` and prefer to manually add the files with `git add` to avoid this issue.
2. It is important to properly describe your PR: what does it do? how? why? is it related to an open ticket in GitHub? It is important for two reasons: 
    - people maintaining open-source software often do it on their free time
    - people using the software needs to know how it works but also why some changes were made - especially that in the case of `pandas-profiling` it is used by thousands of people -.
3. Before opening a PR, check that you follow the [guidelines](https://pandas-profiling.ydata.ai/docs/master/pages/support_contrib/contribution_guidelines.html). This would avoid the linter issue.

aquemy:
After discussion, this new parameters value should be `auto` because we do determine automatically the proper and best metric to use for each pair of columns.

@JamesTook could you update it as well as the documentation to reflect this please?

aquemy:
I asked @JamesTook to rebase interactively his history to avoid multiple noisy commits. By doing this he will not be able to push and I don't want him to force push. Therefore, there is a new PR associated to this: https://github.com/ydataai/pandas-profiling/pull/1092

I close this one as comments have been taken into account.

PR #1088, opened 2022-10-02
Added a Aviation data and script 

agasheaditya:
Example of aviation data frame analysis using pandas profiling package.


akx:
I don't think the repo needs an extra 14 megabyte example XLSX file...

PR #1095, opened 2022-10-05
feat: introduce auto parameter for correlations 

jtook:
The purpose of this feature is to automatically choose the most suitable 'correlation'/'association' metric for each pair of columns .The auto setting is an easily interpretable pairwise column metric that uses previously implemented metrics of the following mapping:

```
vartype-vartype             : method, 
categorical-categorical : Cramer's V, 
numerical-categorical   : Cramer's V (using a discretized numerical column), 
numerical-numerical     : Spearman's ρ. 
```

### DIFF
--- a/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.alerts.rst
+++ b/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.alerts.rst
@@ -23,7 +23,6 @@
       numeric_alerts
       skewness_alert
       supported_alerts
-      timeseries_alerts
       type_date_alert
       unsupported_alerts
    
--- a/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.alerts.rst
+++ b/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.alerts.rst
@@ -23,6 +23,7 @@
       numeric_alerts
       skewness_alert
       supported_alerts
+      timeseries_alerts
       type_date_alert
       unsupported_alerts
    
--- a/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.summary_algorithms.rst
+++ b/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.summary_algorithms.rst
@@ -24,6 +24,7 @@
       describe_numeric_1d
       describe_path_1d
       describe_supported
+      describe_timeseries_1d
       describe_url_1d
       func_nullable_series_contains
       histogram_compute
--- a/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.summary_algorithms.rst
+++ b/docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.summary_algorithms.rst
@@ -24,7 +24,6 @@
       describe_numeric_1d
       describe_path_1d
       describe_supported
-      describe_timeseries_1d
       describe_url_1d
       func_nullable_series_contains
       histogram_compute
--- a/src/pandas_profiling/config_default.yaml
+++ b/src/pandas_profiling/config_default.yaml
@@ -110,6 +110,10 @@ correlations:
       calculate: true
       warn_high_correlations: true
       threshold: 0.9
+    auto:
+    calculate: true
+    warn_high_correlations: true
+    threshold: 0.9
 
 
 # Bivariate / Pairwise relations
--- a/src/pandas_profiling/config_default.yaml
+++ b/src/pandas_profiling/config_default.yaml
@@ -111,9 +111,9 @@ correlations:
       warn_high_correlations: true
       threshold: 0.9
     auto:
-    calculate: true
-    warn_high_correlations: true
-    threshold: 0.9
+      calculate: true
+      warn_high_correlations: true
+      threshold: 0.9
 
 
 # Bivariate / Pairwise relations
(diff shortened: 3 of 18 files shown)


aquemy (approved):

fabclmnt (changes requested):
Don't forget to add to the documentation. 

eg. https://pandas-profiling.ydata.ai/docs/master/pages/reference/api/_autosummary/pandas_profiling.model.correlations.html?highlight=correlation#module-pandas_profiling.model.correlations

https://pandas-profiling.ydata.ai/docs/master/pages/advanced_usage/corr_mat_access.html?highlight=correlation

docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.summary_algorithms.rst
```diff
@@ -24,6 +24,7 @@
       describe_numeric_1d
       describe_path_1d
       describe_supported
+      describe_timeseries_1d
```
fabclmnt, line 27: This seems not to belong to your PR

docsrc/source/pages/reference/api/_autosummary/pandas_profiling.model.alerts.rst
```diff
@@ -23,6 +23,7 @@
       numeric_alerts
       skewness_alert
       supported_alerts
+      timeseries_alerts
```
fabclmnt, line 26: This seems not to belong to your PR

sbrugman (changes requested):
See other comment

fabclmnt:
> Had to look into the previous PRs to see that this is superseding the previous ones. @jtook It would be better to create a feature request issue where we can discuss the broader goal, naming etc. of the feature, so that the PR can be linked for implementation.
> 
> For instance, the name `auto` could be confusing as there is also [autocorrelation](https://en.wikipedia.org/wiki/Autocorrelation). We'd rather not have a discussion for optimal name here. Similarly, we can document there the pro's and cons for adding this method (PhiK is also able to provide coefficients for mixed-type variables).
> 
> On the implementation side: for performance it would make sense to reuse the correlation coefficients that have been already computed for categorical-categorical and numerical-numerical (if enabled).

Although I do agree that's important, we are still setting the best practices for the flows, and given that the PR has already a small description of the feature it will be rather easy to follow up on the progress. 

I think `auto` is not confusing with autocorrelation given the way the configuration file is architected. This was discussed and for now, we will move forward with `auto` - which is a decision in line with what we see in packages.

Regarding the documentation, I do think we can have it improved, and add to the documentation an extensive understanding of the differences between the different association metrics (as it is missing for all of them anyway). My suggestion is to open a docs issue with that detail and address it separately.

fabclmnt (approved):

src/pandas_profiling/config_default.yaml
```diff
@@ -110,6 +110,10 @@ correlations:
       calculate: true
       warn_high_correlations: true
       threshold: 0.9
+    auto:
+    calculate: true
```
sbrugman, line 114: The indentation seems off here (in config_minimal.yaml it is correct)

src/pandas_profiling/model/pandas/correlations_pandas.py
```diff
@@ -138,3 +151,56 @@ def pandas_phik_compute(
...
 
     return correlation
+
+
+@Auto.compute.register(Settings, pd.DataFrame, dict)
+def pandas_auto_compute(
+    config: Settings,
+    df: pd.DataFrame,
+    summary: dict,
+    n_bins: int = 10,
```
sbrugman, line 161: Could we document how the user can change the `n_bins` argument?
(Personally I'd add it as an option under `auto`, just below `threshold` in `config.yml`, with an additional field added to the `Correlation` class in `config.py`)

Does the overload on `Auto.compute` still work when passing `n_bins`?

jtook:
> Had to look into the previous PRs to see that this is superseding the previous ones. @jtook It would be better to create a feature request issue where we can discuss the broader goal, naming etc. of the feature, so that the PR can be linked for implementation.
> 
> For instance, the name `auto` could be confusing as there is also [autocorrelation](https://en.wikipedia.org/wiki/Autocorrelation). We'd rather not have a discussion for optimal name here. Similarly, we can document there the pro's and con's for adding this method (PhiK is also able to provide coefficients for mixed type variables).
> 
> On the implementation side: for performance it would make sense to reuse the correlation coefficients that have been already computed for categorical-categorical and numerical-numerical (if enabled).

The idea behind this PR is that the ‘auto’ correlation should be the default setting (with all the other correlations disabled) for assessing correlations. If the user wants more control over calculating the correlations, they should disable the 'auto' correlation and choose the correlation metric/s that better suits their use case.

src/pandas_profiling/model/pandas/correlations_pandas.py
jtook: 1. I have taken your approach and implemented this as you have recommended. 
2. I have tested calling the pandas_auto_compute function multiple times with/without n_bins parameter without failure. For instance it passes with the following example:

```
auto_result = pandas_auto_compute(test_config, df, summary, n_bins = 12)
auto_result = pandas_auto_compute(test_config, df, summary)
auto_result = pandas_auto_compute(test_config, df, summary, n_bins = 10)

```

Let me know if this is what you mean!

aquemy:
Hi @sbrugman,

I don't think there is any blocker on this but as we want to release asap and we still need to test more, I will merge this.

PR #1096, opened 2022-10-05
Fixes #638 : Added filter to browse through columns easily

AlphaGaurav7:
I have added an input box functionality which can filter the columns for easier access to the desired column. 

image

PR #1101, opened 2022-10-06
feat: added filter to locate columns

### LINKED ISSUE #638: Provide ways to quickly locate a column in the HTML report
**Missing functionality**

In the situation of profiling large data frames (hundreds of columns), it is very hard to locate a specific column. It would be great if the HTML report provide a way for this. For example, provide a filtering box. When a user type in something in the filtering box, show only match columns. 

**Proposed feature**

Provide a filtering box in the HTML report. When a user type in something in the filtering box, show only match columns. 

**Alternatives considered**

**Additional context**

g-kabra:
This is a follow-up PR to the PR made earlier (#1096). Closes #638 
Have changed the input from an text field to a dropdown as per @fabclmnt's suggestion.

Here's how it looks and works now:

https://user-images.githubusercontent.com/57868024/194428807-a7642deb-6ba5-4404-95ef-3e9605ba10cd.mp4

The dropdown isn't visible due to restrictions on the screen-recorder, here's an image of it in action for reference.

image

P.S. I'm sorry for the hassle in the previous PR, I haven't worked with git very much. Thank you for your patience.


.pre-commit-config.yaml
```diff
@@ -6,7 +6,7 @@ repos:
     rev: 22.3.0
     hooks:
     - id: black
-      language_version: python3.10
+      language_version: python3.8
```
fabclmnt, line 9: @AlphaGaurav7 I don't think this is a change required for this development. Can you please have it reverted?

fabclmnt (changes requested):
Thank you for the changes :) 

From my side only the previous comment.

src/pandas_profiling/report/structure/report.py
```diff
@@ -58,7 +59,15 @@ def render_variables_section(config: Settings, dataframe_summary: dict) -> list:
...
-    templs = []
+    templs: List[Renderable] = []
+    items = []
+
+    for idx, summary in dataframe_summary["variables"].items():
+        items.append(idx)
+
+    dropdown = Dropdown(name="variables-dropdown", id="variables-dropdown", items=items)
+
+    templs.append(dropdown)
```
akx, line 70: I think this shortens to just

```suggestion
    dropdown = Dropdown(
        name="variables-dropdown",
        id="variables-dropdown",
        items=list(dataframe_summary["variables"]),
    )
    templs.append(dropdown)
```
since you're iterating over what seems to be a dict's `items` just to get the key and push it to a list; you can iterate over keys with just `for key in dataframe_summary["variables"]`, and from that you can just use `list()`. Then, the `items` local doesn't seem necessary.

akx (commented):
Some suggestions!

AlphaGaurav7:
All changes are valid and great suggestions! Gives me an insight on how to be more consistent and concise with the codebase. Thank you so much.

fabclmnt (approved):
All good 🚀 
Thank you @AlphaGaurav7 for your contribution and @akx for the great suggestions.

AlphaGaurav7:
It's all my pleasure! Looking forward to contributing more :)

sbrugman:
@g-kabra Nice work. It should now be easy to create the same functionality for ipywidgets as well (see [docs](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Selection-widgets)).
(This makes the behaviour consistent accross html/jupyter)

g-kabra:
@sbrugman I'll give it a go for sure! I'll let you know if I run into any problems.

g-kabra:
Any hint as to how can I access the variable widgets? Like a selector that I could use in jquery?

sbrugman:
@g-kabra A possible direction (haven't thought too much about it, might be other ways to achieve the same through event handlers)

Currently the dropdown is on the same `renderable` level as the variables. For ipywidgets a slightly different approach might be simpler, where `Dropdown` wraps the variables (and to make it consistent this should also happen for HTML). On the callback then you could toggle the visibility of the child elements, similar to:

https://github.com/ydataai/pandas-profiling/blob/develop/src/pandas_profiling/report/presentation/flavours/widget/collapse.py

g-kabra:
Hey @sbrugman ! I've looked into the problem for a while, and I think if I try wrapping the "collapse" items into the dropdown, it should work well and good for ipywidgets, but doing the same in the HTML would make the variables only accesible through the dropdown. Is that the intended functionality we are expecting? If not, I'll implement the wrapping for ipywidgets, and try making the dropdown as a container for the variables, and leave the HTML as is.

sbrugman:
@g-kabra I think we're on the right track. In `report.py` line 360, you could place the container as `item` on the dropdown (`items` might then not even be needed, but we'll look into that later).


```python
if len(summary["variables"]) > 0:
    section_items.append(
        Dropdown(
            name="variables-dropdown",
            id="variables-dropdown",
            items=list(dataframe_summary["variables"]),
            item=Container(
              render_variables_section(config, summary),
              sequence_type="accordion",
              name="Variables",
              anchor_id="variables",
          )
      )
  )
```

To ensure that the variable section is rendered, the dropdown template needs to be extended with (`src/pandas_profiling/report/presentation/flavours/html/templates/dropdown.html`):

```
{{ item.render() }}
```
(see `footer.html` as example)

For HTML, this shouldn't change anything to the functionality you've built so far. For ipywidgets however, you've got access to the `item` attribute withing `Dropdown` that is needed to open/close or show/hide the widgets. 

Does this help you further?

g-kabra:
This definitely clear things up! I'll get going and implement these changes! Thank you for the guidance.

g-kabra:
Hey @sbrugman ! I've tried making this work and I have succesfully wrapped the Variables within the dropdown and am able to access the variables within ipywidgets as well. The only issue is that I can hide/unhide data within the accordion, but not the tab for it which seems off. So I've tried implementing it so that it expands the tab which is chosen within the drop down. Video for reference:

https://user-images.githubusercontent.com/57868024/194849953-631da108-a1c0-437e-80a8-1db9d7b42e14.mp4

Any tips on how to hide unhide the elements themself? Also, the HTML side is now functional but needs some design work implemented.

image

fabclmnt:
@g-kabra feel free to have it merged !! 

Thank you for your contribution 🚀

g-kabra:
I have pushed the last set of changes from my end! It was great working on this issue :)

aquemy:
Awesome feature @g-kabra !
It is going to be release with the next version.

aquemy:
Hi @g-kabra

I had to revert because we noticed that the behavior was inconsistent on some datasets.
Could you re-open a PR with your current state of work please?
We will then provide more details on what is actually failing to obtain a more robust solution.

g-kabra:
Sure @aquemy . Let's work on the errors. #1115 up and ready for work!

PR #1109, opened 2022-10-13
fix: cramer's correlation fails with missing values

### LINKED ISSUE #1099: Cramer correlation matrix is not computed
/usr/local/lib/python3.7/dist-packages/pandas_profiling/model/correlations.py:61: UserWarning:There was an attempt to calculate the cramers correlation, but this failed.
To hide this warning, disable the calculation
(using `df.profile_report(correlations={"cramers": {"calculate": False}})`
If this is problematic for your use case, please report this as an issue:
https://github.com/ydataai/pandas-profiling/issues
(include the error message: 'No data; `observed` has size 0.')
 pls help. My dataset is https://busan302.mycourses.work/data/house_price_train.csv

alexbarros:
Fixes #1099.
Cramer's correlation fails when the contingency table between two columns is empty, this seems to be a frequent issue in cases where the columns have a high percentage of missing values.

### DIFF
--- a/src/pandas_profiling/model/pandas/correlations_pandas.py
+++ b/src/pandas_profiling/model/pandas/correlations_pandas.py
@@ -97,9 +97,12 @@ def pandas_cramers_compute(
 
     for name1, name2 in itertools.combinations(categoricals, 2):
         confusion_matrix = pd.crosstab(df[name1], df[name2])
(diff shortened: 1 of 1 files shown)


fabclmnt (approved):
Lgtm

PR #1112, opened 2022-10-18
chore(actions):  remove manual source code versioning

vascoalramos:
This PR adds the final step of Release CI Automation: automatic versioning on the python package.

As such, the version.py was removed from the source code and the development installation instructions were updated.

Additional steps (from issues found during last release):
- fix push permissions
- update master folder on `gh-pages` branch instead of creating new tag branch
- fix changelog creation and use md instead of rst on new changelogs


portellaa (approved):

sbrugman:
@vascoalramos `__version__` is still present in `__init__.py`: https://github.com/ydataai/pandas-profiling/blob/chore/ciAutomation/src/pandas_profiling/__init__.py

This is an useful feature when for users that would like to obtain the current version, simply:
```
from pandas_profiling import __version__
print(__version__)
```

I understand removing the version writing in `setup.py`. Obtaining the version information via conda/pip may in practise be troublesome for users with multiple python environments. This constant is guaranteed to be coming from the same version as the current python environment. I'd suggest keeping the `__version__` feature, but optimizing the build.

(replacing `setup.py` + `manifest.in` + `requirements.txt` with `pyproject.toml` would be a great next step too)

vascoalramos:
> Great work! (Overlooked a change in previous post)
> 
> (replacing `setup.py` + `manifest.in` + `requirements.txt` with `pyproject.toml` would be a great next step too)

thank you. i will address that in another pr

sbrugman (approved):

portellaa (approved):

PR #1115, opened 2022-10-19
feat: added filter to locate columns

g-kabra:
Follow up PR to correct inconsistencies encountered in previous PR for same feature (#1101).


fabclmnt:
@g-kabra after merging into `develop` I was able to identify some odd behavior in the filter:

- For some datasets, the filter was not able to locate the selected variable, which means the variables section would just "disappear" if a column was selected;
- The sections order was altered - interactions are being shown before the variables section for profiles where the minimal setting is set to False.

Uploading Screen Recording 2022-10-19 at 12.07.45 PM.mov…

fabclmnt:
Here the report for the titanic dataset
https://user-images.githubusercontent.com/30911746/196782652-1c20ef9c-e6e4-4aed-838f-df053007a883.mov

g-kabra:
I did find one issue in the jQuery function I had setup. Since the names were being converted to lowercase, a match wasn't being found. I have updated that and will be committing. Reviews and suggestions appreciated.

g-kabra:
As far as the order of Interaction is concerned, I'll sort it out and commit the same.

g-kabra:
These two issues are solved. Sorry I had overlooked these in a hurry.

aquemy:
I checked the code from @g-kabra locally directly from his fork and it works for Titanic now.

In general, the position of the select is a bit awkward because it breaks the HTML border, but this is a cosmetic thing we might fix later.

aquemy (approved):

PR #1119, opened 2022-10-20
fix: improve description and correct plot for ‘auto’ correlation

jtook:


aquemy:
@fabclmnt the other examples are in python while this one is a Notebook. I am OK with that, and the example if fine to me but I just want to check with you that we can allow Notebooks as examples in the future.

aquemy (approved):

fabclmnt (dismissed):
@jtook this is not a `feat` it is at maximum a `fix` because you included changes at the level of the `correlation.py` file. 

As examples are out of the package they should not be considered for version increment of the package. Can you please change the PR message in accordance?
Also, can you please confirm the license of the dataset used? 

@aquemy as this is for the feature folder and not for the use-cases folder, I'll ask to have this changed to a python file. The change is minor and it eases the consumption of the code to generate automated documentation, as we have in other examples.

Notebooks should be used for examples such as `usaairquality` or `employees-chicago`.

jtook:
>The changes you requested have been made. I used a dataset that is already being used in the examples folder from the [UCI dataset collection](https://archive.ics.uci.edu/ml/datasets/bank+marketing )

PR #1133, opened 2022-10-31
feat: spark profiling

chanedwin:
This PR includes the initial support for Profiling Spark DataFrames


src/pandas_profiling/model/describe.py
```diff
@@ -154,7 +155,7 @@ def describe(
 
         pbar.set_postfix_str("Get reproduction details")
         package = {
-            "pandas_profiling_version": __version__,
+            "pandas_profiling_version": "3.1",
```
aquemy, line 158: Shouldn't it be `__version__`?

src/pandas_profiling/model/correlations.py
```diff
@@ -89,10 +89,6 @@ def calculate_correlation(
     Returns:
         The correlation matrices for the given correlation measures. Return None if correlation is empty.
     """
-
-    if len(df) == 0:
```
aquemy, line 93: is it in purpose or a wrong merge?

src/pandas_profiling/model/describe.py
```diff
@@ -50,9 +49,6 @@ def describe(
             - package: package details.
     """
 
-    if df is None:
```
aquemy, line 53: is it in purpose or a wrong merge?

src/pandas_profiling/profile_report.py
```diff
@@ -4,13 +4,20 @@
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
+try:
```
aquemy, line 7: Wow, very nice code! 8)

src/pandas_profiling/profile_report.py
```diff
@@ -124,17 +128,14 @@ def __init__(
                 .update(report_config.dict(exclude_defaults=True))
             )
 
-        if config is not None:
```
aquemy, line 127: I suspect this to be a wrong merge as it looks like a recent fix for the report config. @alexbarros FYI

aquemy (changes requested):
A couple of minor comments, in particular because I am not sure if this is due to a wrong merge or intended.

src/pandas_profiling/__init__.py
```diff
@@ -2,19 +2,22 @@
 
 .. include:: ../../README.md
 """
+import importlib.util
 
 from pandas_profiling.compare_reports import compare
 from pandas_profiling.controller import pandas_decorator
 from pandas_profiling.profile_report import ProfileReport
-from pandas_profiling.version import __version__
```
alexbarros, line 9: why is this being removed?

src/pandas_profiling/config.py
```diff
@@ -305,11 +306,22 @@ class Config:
...
+        "cramers": Correlation(key="cramers"),
+        "kendall": Correlation(key="kendall"),
     }
 
-    correlation_table: bool = True
+    correlations["auto"].calculate = True
+    correlations["pearson"].calculate = False
+    correlations["phi_k"].calculate = False
+    correlations["cramers"].calculate = False
+    correlations["kendall"].calculate = False
```
alexbarros, line 324: you can just do `"spearman": Correlation(key="spearman", calculate=True)`

src/pandas_profiling/model/spark/dataframe_spark.py
```diff
@@ -0,0 +1,54 @@
...
+
+from pyspark.sql import DataFrame
+
+from pandas_profiling.config import Settings
+from pandas_profiling.model.dataframe import check_dataframe, preprocess
+
+
+@check_dataframe.register
+def spark_check_dataframe(df: DataFrame) -> None:
+    # FIXME: never...
```
alexbarros, line 11: it its not a useful comment we should remove it

src/pandas_profiling/profile_report.py
alexbarros: I believe it was done intentionally since this check was moved a bit upwards. Still, it can have some unexpected behaviors because the parameters `explorative`, `sensitive`, `dark_mode`, and `orange_mode` will be applied over a predefined config, which may lead to overwriting the configuration.

src/pandas_profiling/report/structure/variables/render_categorical.py
```diff
@@ -403,11 +403,13 @@ def render_categorical(config: Settings, summary: dict) -> dict:
...
 
-    if length:
+    # length isn't being computed for categorical in spark
+    if length and "max_length" in summary:
         length_table, length_histo = render_categorical_length(config, summary, varid)
         overview_items.append(length_table)
 
-    if characters:
+    # characters isn't being computed for categorical in spark
+    if characters and "category_alias_counts" in summary:
```
alexbarros, line 412: will `category_alias_counts` be present for pandas?

src/pandas_profiling/report/structure/variables/render_categorical.py
```diff
@@ -403,11 +403,13 @@ def render_categorical(config: Settings, summary: dict) -> dict:
 
     overview_items = []
 
-    if length:
+    # length isn't being computed for categorical in spark
+    if length and "max_length" in summary:
```
alexbarros, line 407: will `max_length` be present for pandas?

src/pandas_profiling/model/correlations.py
fabclmnt: this breaks spark integration, so it was removed from here and place changed. 

Spark Dataframe has no len method

src/pandas_profiling/model/describe.py
fabclmnt: Logic not needed as this method is not user interfaced. You already have this validation done earlier, too many duplicate and hard to maintain validation rules as they are spread across files.

src/pandas_profiling/profile_report.py
fabclmnt: As Barros mentioned, it is intended as previous installations were simply ignoring those inputs. The change of code fix this.

src/pandas_profiling/__init__.py
fabclmnt: Shouldn't it was a mistake. Fixed now.

src/pandas_profiling/config.py
fabclmnt: I was following what was done previously and kept the same logic so I don't need to change both (check the history for the PR).

src/pandas_profiling/report/structure/variables/render_categorical.py
fabclmnt: It is already and working for both.

src/pandas_profiling/report/structure/variables/render_categorical.py
fabclmnt: Yes, working for pandas.

src/pandas_profiling/profile_report.py
```diff
@@ -147,11 +147,46 @@ def __init__(
...
+                    "DataFrame is empty. Please" "provide a non-empty DataFrame."
+                )
+
+    @staticmethod
+    def __initialize_config(config_file: Optional[Union[Path, str]]):
+        # Ainda validar se vamos avançar
+        return
+
+    @staticmethod
+    def __initialize_dataframe(df: [pd.DataFrame, sDataFrame], report_config: Settings):
```
aquemy, line 186: It is missing `Union`, should be:
```python
def __initialize_dataframe(df: Union[pd.DataFrame, sDataFrame], report_config: Settings):
```

fabclmnt: Indeed! Nice catch.

aquemy (approved):

PR #1140, opened 2022-11-09
refactor: replace missingno

alexbarros:
Missingno has seen very few updates in the last 2 years, and to avoid potential dependencies' conflicts and vulnerabilities a version of missingno was incorporated.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,8 +9,6 @@ numpy>=1.16.0,<1.24
 # Could be optional
 # Related to HTML report
 htmlmin==0.1.12
-# Missing values
-missingno>=0.4.2, <0.6
 # Correlations
 phik>=0.11.1,<0.13
 # Examples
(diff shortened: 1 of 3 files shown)


portellaa (commented):
A few considerations on this,

We haven't be using refactor a lot, although in this case it makes sense, but since this makes sense, let's make this a patch, otherwise it will not have impact.

so in the `.releaserc.json` file, change this to:
```json
{
  "plugins": [
    ["@semantic-release/commit-analyzer", {
      "preset": "angular",
      "releaseRules": [
        {"type": "chore", "release": "patch"},
        {"type": "refactor", "release": "patch"}
      ]
    }],
    "@semantic-release/release-notes-generator",
    "@semantic-release/github"
  ]
}

otherwise this will not produce a release.

In another topic, i would suggest to instead of move the code into the `pandas_profiling` repo, we could fork the `missingno` project and upload this to pypi under our name or pupose them to change this somehow.

But this is just my two cents for this

PR #1158, opened 2022-11-21
fix: update config files to only calculate 'auto' correlation

jtook:
The Auto correlation matrix calculation should be enabled by default, with the other correlation matrices disabled. 

The example file for this feature has been updated to show:

- How to disable the 'auto' correlation matrix calculation.
- How to update the n_bins setting.
- How to enable the calculation of other correlation matrices.

### DIFF
--- a/src/pandas_profiling/config_default.yaml
+++ b/src/pandas_profiling/config_default.yaml
@@ -90,23 +90,23 @@ missing_diagrams:
 
 correlations:
     pearson:
-      calculate: true
+      calculate: false
       warn_high_correlations: true
       threshold: 0.9
     spearman:
-      calculate: true
+      calculate: false
       warn_high_correlations: false
       threshold: 0.9
     kendall:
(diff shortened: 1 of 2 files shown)


aquemy (approved):

fabclmnt (dismissed):
Some changes are required. Please confirm with @alexbarros regarding the configuration file for the time-series (tsmode).

src/pandas_profiling/config_default.yaml
```diff
@@ -90,23 +90,23 @@ missing_diagrams:
 
 correlations:
```
fabclmnt, line 91: @jtook is this configuration also reflected for the Tsmode?

jtook: Yes, so this is reflected when tsmode is enabled and the default configuration file is specified

aquemy (approved):

PR #1176, opened 2022-11-26
feat: detect timeseries dataset automatically

rivanfebrian123:
If `tsmode` in `ProfileReport` is unspecified, it will try to guess whether the dataset is a time series. This should make this library more beginner friendly.

### DIFF
--- a/src/pandas_profiling/profile_report.py
+++ b/src/pandas_profiling/profile_report.py
@@ -2,7 +2,7 @@ import copy
 import json
 import warnings
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, Tuple
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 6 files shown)


src/pandas_profiling/profile_report.py
```diff
@@ -55,7 +56,7 @@ def __init__(
         sensitive: bool = False,
         dark_mode: bool = False,
         orange_mode: bool = False,
-        tsmode: bool = False,
```
fabclmnt, line 58: `tsmode` should not be an optional input. Please keep it as _False_. This was a requirement when the feature was implemented.

src/pandas_profiling/profile_report.py
```diff
@@ -131,9 +132,11 @@ def __init__(
         if kwargs:
             report_config = report_config.update(kwargs)
 
+        if tsmode is None:
```
fabclmnt, line 135: instead of None use the boolean to run or not this validation.

fabclmnt (changes requested):
@rivanfebrian123 Can you please explain and detail what do you expect to achieve with this PR?

Based on the changes, it seems like this is prone to create reports with _False_ positives that are not expected.  In case, the objective is to automate the generation of time-series report, unfortunately, I'll have to reject the PR as the current automated detection is not yet ready to serve properly the automated selection between a regular and time-series profile.

PR #1177, opened 2022-11-26
fix(utils): use 'urllib.request' instead of 'requests'

rivanfebrian123:
`requests` doesn't seem to handle TLS/SSL pretty well. It returns `Forbidden` when we try to get a content from an HTTPS URL such as `https://query1.finance.yahoo.com/v7/finance/download/BABA?period1=1635724800&period2=1667260800&interval=1d&events=history&includeAdjustedClose=true`. That's why we should use `urllib.request` instead.

On the other hand, I need to remove `raise FileNotFoundError("Could not download resource")` as `urllib.request.urlopen` already handle that with its `HTTPError`.

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,1 @@
+titleAndCommits: true
--- a/.github/workflows/merge-master.yml
+++ b/.github/workflows/merge-master.yml
@@ -40,7 +40,9 @@ jobs:
       id: version
       run: echo "::set-output name=value::${{ steps.semantic.outputs.new_release_version }}.rc${{ steps.build_number.outputs.value }}"
 
+
   docs-changelog:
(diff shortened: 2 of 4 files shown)


fabclmnt (approved):

fabclmnt (approved):

aquemy:
As requests is not used, it should be removed from `requirements.txt`:
https://github.com/ydataai/pandas-profiling/blob/5e75fd275d14c8ce7ba49d0a15ec26810c4c0e73/requirements.txt#L15

Also, there are two files to update then:
https://github.com/ydataai/pandas-profiling/blob/develop/tests/issues/test_issue377.py
https://github.com/ydataai/pandas-profiling/blob/develop/tests/issues/test_issue537.py

rivanfebrian123:
@aquemy Sorry, haven't take a look at those

PR #1179, opened 2022-11-27
chore(examples): working practical cloud-platform notebook examples

rivanfebrian123:
The current Google Colab and Binder notebook examples don't work quiet well. Therefore, I add some working examples for those. BTW, there is no need to restart kernel whatsoever.


fabclmnt:
@rivanfebrian123 thank for contributing! 
Can you please change your PR request to `dev` branch instead of master?

fabclmnt (changes requested):
Instead of creating new notebooks for cloud specifically, may I suggest to update the existing notebooks to run both locally and in Google Colab?

rivanfebrian123:
Would you mind if I exclude `profile.to_widgets()` from the notebooks as it return `TypeError: type of the return value must be pandas_profiling.report.presentation.core.renderable.Renderable; got ipywidgets.widgets.widget_box.VBox instead` on both Google Colab and Binder?

I mean, If we have separated the notebook files, localhost users can take advantage of `profile.to_widgets()` where they can't do that on online platforms.

fabclmnt:
> Would you mind if I exclude `profile.to_widgets()` from the notebooks as it return `TypeError: type of the return value must be pandas_profiling.report.presentation.core.renderable.Renderable; got ipywidgets.widgets.widget_box.VBox instead` on both Google Colab and Binder?
> 
> I mean, If we have separated the notebook files, localhost users can take advantage of `profile.to_widgets()` where they can't do that on online platforms.

Hi @rivanfebrian123 Got your point! 
Let's keep them separately as per your initial suggestion  :)

smith-kyle:
For PRs like this where you're making changes to Jupyter Notebooks, if you want the same code review experience as other GitHub file types you can use [GitNotebooks](https://gitnotebooks.com/)

image
 
It's free! (I'm the solo dev, looking for early users)

PR #1180, opened 2022-11-28
 feat: detect timeseries dataset automatically

rivanfebrian123:
**Motivation:**
When I first know this library, I didn't know if it has timeseries mode. I was absolutely beginner in data science.

With this PR, I hope it can give practical output for newcomers, as they might don't know the difference between timeseries dataset and the regular ones.

**Replies for previous PR:**

> The automation will return too many false positives, which is not ideal. Please only activate the evaluation in case it is specifically requested by the user to generate a time-series report.

Thanks for noticing it. `dateutils.parser.parse` indeed matches broad types of string which is prone to false positivity. That's why I changed it to a more visible regex that matches only common datetime formats.

Regarding to false positivity, I already checked it in my notebook: https://colab.research.google.com/drive/1jCQFblAoNWtY2Zp26kwzdGzYvwlr1Mta?usp=sharing

There you can see that the only timeseries dataset is "BABA.csv". You can also try some other timeseries and regular datasets there.

> `tsmode` should not be an optional input. Please keep it as False. This was a requirement when the feature was implemented.

I think it should be fine as we can't always assume that a dataset is non timeseries and the feature is relatively new.

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,1 @@
+titleAndCommits: true
--- a/.github/workflows/merge-master.yml
+++ b/.github/workflows/merge-master.yml
@@ -40,7 +40,9 @@ jobs:
       id: version
       run: echo "::set-output name=value::${{ steps.semantic.outputs.new_release_version }}.rc${{ steps.build_number.outputs.value }}"
 
+
   docs-changelog:
+    if: needs.prepare.outputs.new_release == 'true'
     name: Update changelog on docs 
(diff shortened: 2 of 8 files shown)


fabclmnt:
@rivanfebrian123 as I've mentioned in the last PR, I understand the motivation behind it, but the proposed solution won't solve the issue. 

Time-series datasets many times have no DateTime involved. Can you please open and issue instead and discuss the solution prior the implementation, that way we can align best.

PR #1195, opened 2022-12-05
fix: do no raise exception for percentage formatter

aquemy:
A percentage might be higher than 100% when it expresses a ratio for instance.
We should not control this behavior at the formatter level.


aquemy:
The issue happened when the two datasets are not the same size. In particular, when the first one is larger.
The root of the problem is that the frequency table then uses the number of rows of the first report for the second report which is correct.

All the other information are properly displayed and uses the size of their respective dataset:
image

fabclmnt (approved):
Nice catch! 🚀 all looking good

PR #1200, opened 2022-12-06
fix: compare report warnings always showing the last alert type

ricardodcpereira:
The initialization of the dictionary used to save the alerts was reusing the same list for every alert type. Since this was a shared reference, the code was updating the same list for all alert types (therefore only keeping the last alert type). The issue is solved (see the image below). All unit tests passed.

Screenshot from 2022-12-06 14-56-23


aquemy (approved):
Damn, nice catch! :champagne:

PR #1203, opened 2022-12-08
feat: implement imbalanced warning

jtook:
**Context**
Users may want to assess how balanced their Categorical and Boolean columns are and be warned if there are signs of high imbalance. This becomes increasingly difficult for them to manually interpret with a higher number of classes (for Categorical Variables). 

**Solution**

This PR aims to introduce an elegant and interpretable score which can be used to alert the user of high imbalance in their categorical or Boolean variables.

The changes to the report:

(alerts)
class_imbalance_alerts

**The Maths**

The class balance score for categorical and boolean variables uses entropy to calculate a  bounded score between 0 and 1. A perfectly uniform distribution would return a score of 0, and a perfectly imbalanced distribution would return a score of 1.

When dealing with probabilities with finite values (e.g categorical), entropy is maximised the ‘flatter’ the distribution is. (Jaynes: Probability Theory, The Logic of Science)

To calculate the class imbalance, we calculate the entropy of that distribution and the maximum possible entropy for that number of classes.
 
To calculate the entropy of the 'distribution' we use value counts (e.g frequency of classes) and we can determine the maximum entropy as log2(number of classes).

We then divide the entropy by the maximum possible entropy to get a value between 0 and 1 which we then subtract from 1.

### DIFF
--- a/docsrc/source/pages/getting_started/overview.rst
+++ b/docsrc/source/pages/getting_started/overview.rst
@@ -47,7 +47,7 @@ For each column, the following information (whenever relevant for the column typ
 The report contains three additional sections: 
 
 * **Overview**: mostly global details about the dataset (number of records, number of variables, overall missigness and duplicates, memory footprint)
-* **Alerts**: a comprehensive and automatic list of potential data quality issues (high correlation, skewness, uniformity, zeros, missing values, constant values, between others) 
+* **Alerts**: a comprehensive and automatic list of potential data quality issues (high correlation, imbalance, skewness, uniformity, zeros, missing values, constant values, between others) 
 * **Reproduction**: technical details about the analysis (time, version and configuration)
 
 The package can be used via code but also directly as a CLI utility. The generated interactive report can be consumed and shared as regular HTML or embedded in an interactive way inside Jupyter Notebooks. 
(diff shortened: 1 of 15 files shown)


docsrc/source/pages/getting_started/overview.rst
```diff
@@ -32,7 +32,7 @@ which automatically generates a standardized univariate and multivariate report
 For each column, the following information (whenever relevant for the column type) is presented in an interactive HTML report:
 
 * **Type inference**: detect the types of columns in a ``DataFrame``
-* **Essentials**: type, unique values, indication of missing values
+* **Essentials**: type, unique values, imbalance, indication of missing values
```
aquemy, line 35: We should not add the imbalance information to the overview in general.

src/pandas_profiling/report/structure/variables/render_boolean.py
```diff
@@ -56,6 +56,11 @@ def render_boolean(config: Settings, summary: dict) -> dict:
                 "value": fmt_percent(summary["p_missing"]),
                 "alert": "p_missing" in summary["alert_fields"],
             },
+            {
```
aquemy, line 59: We don't want to display this information in the rendering.

src/pandas_profiling/report/structure/variables/render_categorical.py
```diff
@@ -370,6 +370,11 @@ def render_categorical(config: Settings, summary: dict) -> dict:
                 "value": fmt_percent(summary["p_missing"]),
                 "alert": "p_missing" in summary["alert_fields"],
             },
+            {
```
aquemy, line 373: We don't want to display this information in the rendering.

aquemy (changes requested):
Very cool feature and implementation.

My only request is to remove the field from the essential section. We only wants to have a warning.

As you already calculated the value for the warning, it can stay in the summary object itself, so this is just about not displaying it.

aquemy (approved):

Issue #915, opened 2022-01-27
Error for series with large integers


Loquats:
**Describe the bug**
Profile report HTML cannot be generated for dataframes with very large integers. It seems like an overflow issue. Would it be possible to fix this? Even if a report for this series cannot be generated, it would be great if pandas-profiling could at least handle it gracefully so that the code does not crash.

**To Reproduce**
```
import pandas as pd
from pandas_profiling import ProfileReport

df = pd.DataFrame({"col": pd.Series([716277643516076032 + i for i in range(100)])})
df_profile = ProfileReport(df)
profile_html = df_profile.to_html()
```

<details>
<summary>Full stack trace</summary>
<p>

```
IndexError: index 10 is out of bounds for axis 0 with size 9
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<command-1661028028735696> in <module>
      4 df = pd.DataFrame({"col": pd.Series([716277643516076032 + i for i in range(100)])})
      5 df_profile = ProfileReport(df)
----> 6 profile_html = df_profile.to_html()
      7 profile_html

/databricks/python/lib/python3.8/site-packages/pandas_profiling/profile_report.py in to_html(self)
    366 
    367         """
--> 368         return self.html
    369 
    370     def to_json(self) -> str:

/databricks/python/lib/python3.8/site-packages/pandas_profiling/profile_report.py in html(self)
    183     def html(self) -> str:
    184         if self._html is None:
--> 185             self._html = self._render_html()
    186         return self._html
    187 

/databricks/python/lib/python3.8/site-packages/pandas_profiling/profile_report.py in _render_html(self)
    285         from pandas_profiling.report.presentation.flavours import HTMLReport
    286 
--> 287         report = self.report
    288 
    289         with tqdm(

/databricks/python/lib/python3.8/site-packages/pandas_profiling/profile_report.py in report(self)
    177     def report(self) -> Root:
    178         if self._report is None:
--> 179             self._report = get_report_structure(self.config, self.description_set)
    180         return self._report
    181 

/databricks/python/lib/python3.8/site-packages/pandas_profiling/profile_report.py in description_set(self)
    159     def description_set(self) -> Dict[str, Any]:
    160         if self._description_set is None:
--> 161             self._description_set = describe_df(
    162                 self.config,
    163                 self.df,

/databricks/python/lib/python3.8/site-packages/pandas_profiling/model/describe.py in describe(config, df, summarizer, typeset, sample)
[...]

PR #1233, opened 2023-01-05
fix: issue#915, Error for large integers in Series

Sohaib90:
This patch fixes issue #915: Error for series with large integers. 

The issue is caused when using the `numpy.histogram` function with large integers which causes unevenly spaced bin edges.
here: 

https://github.com/ydataai/pandas-profiling/blob/5b1abac48ed9ed5a9e7e662be30c913acc3e7a5b/src/pandas_profiling/model/summary_algorithms.py#L39

and here:

https://github.com/ydataai/pandas-profiling/blob/5b1abac48ed9ed5a9e7e662be30c913acc3e7a5b/src/pandas_profiling/model/summary_algorithms.py#L52

This can cause the resulting histogram to be distorted or misleading, as the bin sizes may not be uniform.

To resolve this issue, I used the `numpy.histogram_bin_edges` function to compute the bin edges for the data before passing them to the `numpy.histogram` function. This function allows to specify the number of bins and the range of the data, and will compute the bin edges in a way that ensures they are evenly spaced. This fix does not raise an error as reported in the bug report and successfully generates a report. I have also included a `test_issue915.py ` for testing the generation of the report.


Sohaib90:
@alexbarros you might want to look into this as well if you get the chance :)

src/pandas_profiling/model/summary_algorithms.py
```diff
@@ -36,11 +36,13 @@ def histogram_compute(
     stats = {}
     bins = config.plot.histogram.bins
     bins_arg = "auto" if bins == 0 else min(bins, n_unique)
-    stats[name] = np.histogram(finite_values, bins=bins_arg, weights=weights)
+    bins = np.histogram_bin_edges(finite_values, bins=bins_arg)
+    stats[name] = np.histogram(finite_values, bins=bins, weights=weights)
```
alexbarros, line 40: interesting solution, but still seems to behave a bit weirdly with big numbers
```python
> import numpy as np
> arr = np.array([716277643516076032 + i for i in range(100)])
> bins = np.histogram_bin_edges(arr, bins=5)
> np.histogram(arr, bins=bins)
(array([ 0,  0, 65,  0, 35]),
 array([7.16277644e+17, 7.16277644e+17, 7.16277644e+17, 7.16277644e+17,
        7.16277644e+17, 7.16277644e+17]))
```

src/pandas_profiling/model/summary_algorithms.py
```diff
@@ -36,11 +36,13 @@ def histogram_compute(
...
     bins_arg = "auto" if bins == 0 else min(bins, n_unique)
-    stats[name] = np.histogram(finite_values, bins=bins_arg, weights=weights)
+    bins = np.histogram_bin_edges(finite_values, bins=bins_arg)
+    stats[name] = np.histogram(finite_values, bins=bins, weights=weights)
 
     max_bins = config.plot.histogram.max_bins
     if bins_arg == "auto" and len(stats[name][1]) > max_bins:
-        stats[name] = np.histogram(finite_values, bins=max_bins, weights=None)
+        bins_arg = np.histogram_bin_edges(finite_values, bins=max_bins)
+        stats[name] = np.histogram(finite_values, bins=bins_arg, weights=None)
```
alexbarros, line 45: change the var name from bins_arg to bins to keep consistent

Sohaib90: Done

src/pandas_profiling/model/summary_algorithms.py
Sohaib90: True, it is still not evenly distributed like it should be for smaller numbers. What do you propose here? Leaving `np.histogram_bin_edges` raises an error for larger numbers. Is it better to raise error than have weird behavior?

alexbarros: Thinking from a user's perspective, for me is better to have an error being raised than an incorrect plot. If I know that there was a problem with the large integers I can preprocess that column and run again, but an incorrect result may lead me to an incorrect interpretation of my data distribution.

Sohaib90: Yeah, that is what I was thinking as well. I think I should make the changes so that it leads to raising an error rather than making an incorrect plot, right?

Issue #1104, opened 2022-10-10
Bug Report (Empty column tagged as Unsupported)


joshua-lagasca-novare:
### Current Behaviour

An empty non-object datatype column must be tagged with an 'Empty' alert.

### Expected Behaviour

An empty non-object datatype column returns tagged with an 'Unsupported' alert.

### Data Description

An empty non-object datatype column.

### Code that reproduces the bug

```Python
import pandas as pd
from pandas_profiling import ProfileReport

df = pd.DataFrame(columns=['a','b']).astype(int)
report = ProfileReport(df, title="bug report", minimal=True)
report.to_file("report.html")
```

### pandas-profiling version

v3.3.0

### Dependencies

```Text
pandas==1.3.5
```

### OS

Ubuntu 20.04.5 LTS x86_64

### Checklist

- [X] There is not yet another bug report for this issue in the [issue tracker](https://github.com/ydataai/pandas-profiling/issues)
- [X] The problem is reproducible from this bug report. [This guide](http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) can help to craft a minimal bug report.
- [X] The issue has not been resolved by the entries listed under [Common Issues](https://pandas-profiling.ydata.ai/docs/master/pages/support_contrib/common_issues.html).

Mathanraj-Sharma:
Shall I take this ticket and work on this?

@fabclmnt

fabclmnt:
@Mathanraj-Sharma feel free to submit a PR with a proposed solution!

alexbarros:
hi @Mathanraj-Sharma, are you still working on this one?

alexbarros:
@Sohaib90 you can go for it

Sohaib90:
@alexbarros I think it has already been worked on? Attaching the "Expected behavior" screenshot for reference:

![image](https://user-images.githubusercontent.com/43241257/212885181-d9512988-2121-4786-905b-e07c2e9a15e6.png)

I think you can close this issue?

alexbarros:
@Sohaib90 yeah, I believe it is a non-issue in the end, but I believe we should raise an exception in case we try to generate a report of an empty data frame so the error became more obvious faster.

Sohaib90:
Alright. I will work on that then :)

Sohaib90:
@alexbarros I have created a PR for the issue. Maybe you can look into it :)

PR #1238, opened 2023-01-20
fix: issue#1104 empty dataframe

Sohaib90:
This patch includes a fix for https://github.com/ydataai/pandas-profiling/issues/1104, which I tackle via adding a check for the `DataFrame`. If it is empty, I raise a `ValueError` highlighting that the `DataFrame` is empty. 

For reference: 
https://github.com/Sohaib90/pandas-profiling/blob/a760c5e4be724d2c9823cb1ebed77d2009572d61/src/pandas_profiling/profile_report.py#L88


alexbarros:
@Sohaib90 Plz create change the PR to merge into `develop` branch and not master

Sohaib90:
@alexbarros is it correct now? Apologies for this, since I am still new to this

alexbarros:
No problem, the branch is correct but it seems that you need to rebase your branch first, since it references commits other than yours.

Sohaib90:
Can you tell me how to do that? I think it got messy :(

Sohaib90:
> Can you tell me how to do that? I think it got messy :(

I think now it should be okay?

Sohaib90:
@alexbarros I see there are a lot of failed checks. Is it something that is fixable on my end?

alexbarros:
[Sohaib90](https://github.com/Sohaib90) the issue with the checks has been fixed. We can now see the errors.
the tests need to be updated to expect the raised error
```
FAILED tests/unit/test_pandas/test_dataframe_empty.py::test_empty[test_data0] - ValueError: DataFrame is empty. Please provide a non-empty DataFrame.
FAILED tests/unit/test_pandas/test_dataframe_empty.py::test_empty[test_data1] - ValueError: DataFrame is empty. Please provide a non-empty DataFrame.
FAILED tests/unit/test_pandas/test_dataframe_empty.py::test_empty[test_data2] - ValueError: DataFrame is empty. Please provide a non-empty DataFrame.
```
and there was a lint issue in your commit, I belive is either trailing whitespace or the length of the line.

Sohaib90:
@alexbarros  I have fixed the test mentioned above, and tested it locally as well. 
Moreover, I have also fixed the commit message (which is verified by Semantic PR) as well. I dont know if that is related to the lint running in the tests

let me know what more I can do

tests/unit/test_pandas/test_dataframe_empty.py
```diff
@@ -13,11 +13,13 @@
...
 
-    assert len(description["correlations"]) == 0
-    assert len(description["missing"]) == 0
+        assert len(description["correlations"]) == 0
+        assert len(description["missing"]) == 0
 
-    html = profile.to_html()
-    assert "Dataset is empty" in html
+        html = profile.to_html()
+        assert "Dataset is empty" in html
```
alexbarros, line 25: this part of the code is not necessary anymore you can remove it, and that is it I believe.

Sohaib90: Alright :)

alexbarros:
[Sohaib90](https://github.com/Sohaib90) is almost there, now it's just some linting issues and it's good to go.

ps. forgot to mention that I had to rebase this PR to fix the CI and you should have rebased prior to pushing the changes that now fail on the PR commit test, sorry about that. A rebase to alter the commit message should do it.

alexbarros:
```
⧗   input: fix: Issue#1104 empty DataFrame
✖   subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
```
You can use git [interactive rebase](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#amending-older-or-multiple-commit-messages) to rename the commit message,

alexbarros:
did you run the `pre-commit run --all-files`? it seems that there are still some minor lint issues that the pre-commit should fix automatically.
ps. sometimes it alters other files that you did not work, if this happen just ignore those and commit the changes only in the files that you changed

Sohaib90:
> did you run the `pre-commit run --all-files`? it seems that there are still some minor lint issues that the pre-commit should fix automatically. ps. sometimes it alters other files that you did not work, if this happen just ignore those and commit the changes only in the files that you changed

I run` pre-commit run --all-checks` locally and I get this error there as well 
`black....................................................................Failed
- hook id: black
- files were modified by this hook
 
reformatted tests/unit/test_pandas/test_dataframe_empty.py
reformatted src/pandas_profiling/profile_report.py

All done! ✨ 🍰 ✨
2 files reformatted, 235 files left unchanged.`

alexbarros (approved):

alexbarros:
[Sohaib90](https://github.com/Sohaib90) Thanks and congrats on your first contribution :tada:

jamesmyatt:
I think this PR unfixes my previous issue: https://github.com/ydataai/pandas-profiling/issues/678. I don't want it to throw an exception if the data frame is empty. I still want it to produce a report with lists of columns, etc.

Sohaib90:
@jamesmyatt we discussed this but the maintainer thought that it would be better to generate an exception since there will be no report to generate if the dataframe is empty and it also allows a sanity check before generating the report.

jamesmyatt:
I suppose that there are 3 kinds of empty data frames:

* 0 rows x n columns: report is still interesting since you can list columns, data types, e.g. int, obj, etc.
* m rows x 0 columns: report is empty, but probably still interesting
* 0 rows x 0 columns: report is empty and not interesting - I think it might be OK to throw an exception here, but only here.

More than anything, you can't even force it to create a report for an empty data frame.

Imagine you're looping over data frames (e.g. tables in a database) and creating HTML reports for offline browsing. Some of those frames might be empty so now I have to write logic to catch the ValueError and skip profiling those tables. However, the set of HTML files it now creates will be determined by both while tables there are in the database, but also whether they are empty. So if an HTML file is missing, then there are lots of reasons why that could be: could it be the table was empty? could it be the table has been dropped? could it be some other error? In this situation, if the frame exists and is empty, then I want to create a report in the pandas-profiling style that says "0 rows, n columns" or whatever; not just silence.

An empty data frame is still a valid data frame. Like I said, if I don't want to generate a report for an empty data frame then that's easy. But if I've asked it to create a report for a valid data frame, then I expect it to create a report.

If you really want to do something special with an empty data frame, then you could raise a warning. But like I said, an empty data frame is still a valid data frame.

jamesmyatt:
Also, I don't think this addresses #1104, since I think that issue is related to how empty columns are presented in the report not that a report is created for empty data frames at all.

fabclmnt:
@jamesmyatt This was the decision made regarding empty dataframes. Unfortunately it is a common issue, and without the proper message, users where always reaching out as they were not aware of what was happening. 

For that reason we've re-instated the limitation, as it provides a better and more detailed explanation for the end-user. 

Regarding the use-case you've mentioned, and yes though the dataframe is still a dataframe, the introduction of Spark made this even more problematic. For now the decision was to revert #678, and provide the same functionality for Spark and Pandas.

The example provided by the user in #1104 was an empty dataframe, and this was our suggestion for the one of the issues presented by the user. Does not cover one of the requests, as we believe _"Unsupported_" tag suits the current status of the tool development (again Spark integration is relevant here) and the majority of the cases.

This might change in the future, the evolution of development. We will continue pursuing a better solution in the next few months to solve all the cases. Happy to open a feature request with the information you've provided.

PR #1239, opened 2023-01-21
Update correlations.py

kenf1:
### attempt to fix: cannot import name 'DataError' from 'pandas.core.base'

- successfully able to run `from pandas.errors import DataError` in Python 3.11.x
   - used try/except to avoid introducing issues for other Python ver

PR #1244, opened 2023-01-24
feat: add 3.11.x compatibility

kenf1:
attempt to fix cannot import name 'DataError' from 'pandas.core.base'

- successfully able to run from pandas.errors import DataError in Python 3.11.x
- used try/except to avoid introducing issues for other Python ver

revised from: #1239

### DIFF
--- a/.github/workflows/sonarqube.yaml
+++ b/.github/workflows/sonarqube.yaml
@@ -15,7 +15,7 @@ jobs:
     runs-on: self-hosted
 
     steps:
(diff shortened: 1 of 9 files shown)


aquemy:
@kenf1 thank you for your PR!
I had to open and merge another one to have it rebased properly and also, and mostly, because we had to change few other things to fully support Python3.11.

The PR is here: https://github.com/ydataai/ydata-profiling/pull/1273

The support will be available from the next release (4.1.0).

PR #1245, opened 2023-01-25
chore: deprecate pandas-profiling

vascoalramos:
Ticket: [[PLT-720](https://ydata.atlassian.net/browse/PLT-720)]

[PLT-720]: https://ydata.atlassian.net/browse/PLT-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vascoalramos (commented):
@fabclmnt  should we keep this setup url link in the deprecation release as it is, delete it or modify it so it points to `https://github.com/ydataai/ydata-profiling`?

fabclmnt:
> @fabclmnt should we keep this setup url link in the deprecation release as it is, delete it or modify it so it points to `https://github.com/ydataai/ydata-profiling`?

Modify to point to ydata-profiling

vascoalramos:
> > @fabclmnt should we keep this setup url link in the deprecation release as it is, delete it or modify it so it points to `https://github.com/ydataai/ydata-profiling`?
> 
> Modify to point to ydata-profiling. Also, Vasco don't forget to add here the Readme I've sent you. It is meant to be used in this branch (the one that will serve the release for pp deprecation)

@fabclmnt , already fixed that and also aligned the branch to only include he strictly necessary (as in the approach from sklearn)

PR #1250, opened 2023-01-26
feat: change package naming

vascoalramos:


fabclmnt (approved):

fabclmnt:
@aquemy can we please remove the documentation files from this PR? There is a branch dedicated to that. I'm asking this for a particular reason, - we want to keep some references as `pandas-profiling` as they are not supported in the most recent versions. 

I've done the work of checking all the paths and where should and should not be changed.

fabclmnt (changes requested):
Remove documentation from this PR. 

Use the other branch instead.

PR #1265, opened 2023-02-10
fix: correlation between columns w/o non-null row

alexbarros:
Fixes an error when calculating the correlation between columns where there is no nonnull simultaneously for both columns, i.e.
 when column A is non-null column B is null and vice versa.


src/ydata_profiling/model/pandas/correlations_pandas.py
```diff
@@ -53,6 +53,10 @@ def _cramers_corrected_stat(confusion_matrix: pd.DataFrame, correction: bool) ->
     Returns:
         The Cramer's V corrected stat for the two variables.
     """
+    # handles empty crosstab
+    if confusion_matrix.empty:
```
aquemy, line 57: Can be test that internally and adjust accordingly if required?

aquemy (approved):

PR #1339, opened 2023-05-22
fix: complete alerts information in json

alexbarros:
Add to render_json the same information available in the HTML for alerts.
previously:
```
"alerts": [
    "[HIGH_CORRELATION] alert on column ascending_sequence",
    "[HIGH_CORRELATION] alert on column descending_sequence_with_noise",
    "[ZEROS] alert on column zeros"
]
```
now:
```
"alerts": [
    "[ascending_sequence] is highly overall correlated with [descending_sequence_with_noise]",
    "[descending_sequence_with_noise] is highly overall correlated with [ascending_sequence]",
    "[zeros] has 1000 (100.0%) zeros"
],
```
Fixes https://github.com/ydataai/ydata-profiling/issues/1301

### DIFF
--- a/src/ydata_profiling/model/alerts.py
+++ b/src/ydata_profiling/model/alerts.py
@@ -27,7 +27,6 @@ def fmt_percent(value: float, edge_cases: bool = True) -> str:
 
     return f"{value*100:2.1f}%"
 
-
 @unique
 class AlertType(Enum):
     """Alert types"""
--- a/src/ydata_profiling/model/alerts.py
+++ b/src/ydata_profiling/model/alerts.py
@@ -27,6 +27,7 @@ def fmt_percent(value: float, edge_cases: bool = True) -> str:
 
     return f"{value*100:2.1f}%"
 
+
 @unique
 class AlertType(Enum):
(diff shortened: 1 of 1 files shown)


src/ydata_profiling/model/alerts.py
```diff
@@ -77,16 +95,11 @@ class Alert:
     def __init__(
         self,
         alert_type: AlertType,
-        values: Optional[Dict] = None,
+        values: Optional[Dict] = {},
```
ricardodcpereira, line 98: This is dangerous because of the structure being mutable (https://www.inspiredpython.com/article/watch-out-for-mutable-defaults-in-function-arguments). It would be better to leave it as it was before.

aquemy:
Shouldn't it be about adding also the extra contextual information about each alert like the score, related columns, etc?

alexbarros:
For now, I matched the same information available in the HTML.
The plan was to transform Alert into a dataclass and have a more json like output, but this required too many changes, so I postponed the refactoring needed to make it work but laid the groundwork by creating the subclasses.

PR #1343, opened 2023-05-23
fix: infer all `Numeric` vars as `TimeSeries` when tsmode=True

alexbarros:
Drop the time dependence check for time series type inference


ricardodcpereira (approved):

aquemy (approved):

fabclmnt:
@alexbarros this does not solve the request from the PR. Can we please revert this branch?
https://github.com/ydataai/ydata-profiling/issues/1292

PR #1363, opened 2023-06-16
fix: install system dependency in macOS test

alexbarros:


alexbarros:
It seems that it was a temporary error, since rerunning fixed the issue

PR #1397, opened 2023-07-24
feat: add ts plot to variables' section details

alexbarros:
Added a section with the time-series plot to the details section, and improved the visibility of the plot comparison
image


alexbarros:
> Can we have examples of the plots with the changes?

just added an example to the description

aquemy (approved):

PR #1408, opened 2023-08-01
fix:error for spark report

luyehui:


aquemy:
@luyehui thank you for opening this PR! 

Apologies but we had to force push dev for a minor release. This now creates a conflict with your PR. Could you rebase it such that there are only your intended commits please?

PR #1423, opened 2023-08-10
fix: gap identification and improve gap statistics and visualization

alexbarros:
fixed an issue in which entries of multiple rows with the same timestamp contributed to misidentification of gaps
corrected the gap statistics to consider only the identified gaps instead of all data intervals and improved the wording of the metrics


aquemy (approved):

aquemy:
It is approved but the tests are failing though.

aquemy (approved):
Just one comment for the doc.

aquemy (approved):

PR #1424, opened 2023-08-11
chore(deps): upgrade numpy 1.18

ndesaunais:


martin-kokos:
I'd appreciate this merged. numpy-1.23 was from Dec 2022 and there are already packages requiring numpy-1.24, which is causing version conflicts.

alexbarros:
It seems that there were some misalignment here, since we already support `numpy>=1.16.0,<1.24`.
https://github.com/ydataai/ydata-profiling/blob/develop/requirements.txt#L8
I will close this PR, since the requirement is already satisfied, we may open a new one to increase the upper limit.

Issue #1384, opened 2023-07-16
Bug Report


boris-kogan:
### Current Behaviour

Dataset with categorical features causes memory error even on tiny dataset.

 File "/usr/local/lib/python3.9/dist-packages/ydata_profiling/profile_report.py", line 439, in _render_json
    description = self.description_set
  File "/usr/local/lib/python3.9/dist-packages/typeguard/__init__.py", line 1033, in wrapper
    retval = func(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/ydata_profiling/profile_report.py", line 245, in description_set
    self._description_set = describe_df(
  File "/usr/local/lib/python3.9/dist-packages/ydata_profiling/model/describe.py", line 151, in describe
    metrics, duplicates = progress(get_duplicates, pbar, "Detecting duplicates")(
  File "/usr/local/lib/python3.9/dist-packages/ydata_profiling/utils/progress_bar.py", line 11, in inner
    ret = fn(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/multimethod/__init__.py", line 315, in __call__
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/ydata_profiling/model/pandas/duplicates_pandas.py", line 37, in pandas_get_duplicates
    df[duplicated_rows]
  File "/usr/local/lib/python3.9/dist-packages/pandas/core/groupby/groupby.py", line 2411, in size
    return self._reindex_output(result, fill_value=0)
  File "/usr/local/lib/python3.9/dist-packages/pandas/core/groupby/groupby.py", line 4143, in _reindex_output
    index, _ = MultiIndex.from_product(levels_list, names=names).sortlevel()
  File "/usr/local/lib/python3.9/dist-packages/pandas/core/indexes/multi.py", line 643, in from_product
    codes = cartesian_product(codes)
  File "/usr/local/lib/python3.9/dist-packages/pandas/core/reshape/util.py", line 60, in cartesian_product
    return [
  File "/usr/local/lib/python3.9/dist-packages/pandas/core/reshape/util.py", line 62, in <listcomp>
    np.repeat(x, b[i]),
  File "<__array_function__ internals>", line 180, in repeat
  File "/usr/local/lib/python3.9/dist-packages/numpy/core/fromnumeric.py", line 479, in repeat
    return _wrapfunc(a, 'repeat', repeats, axis=axis)
[...]

PR #1427, opened 2023-08-15
fix: update duplicates_pandas.py

boris-kogan:
Fixing Bug Report #1384
Dataset with categorical features causes memory error even on tiny dataset.


aquemy (approved):

aquemy:
Thank you for reporting the issue and solving it!

PR #1302, opened 2023-03-31
feat: description_wrapper

vorel99:
I created wrapper class for description. Classes are easier to work with for new contributors like myself.
- dataclasses.asdict is used for transforming description to dict (when saving to json)
- for reverse transformation is used dacite from_dict

PR #1443, opened 2023-09-14
feat: refactor var description

vorel99:
This is my second refactoring pull request. First was #1302 
My aim is to replace dictionaries with classes for easier feature developing.

* new dataclasses for series description and series description hashable
* add property for common propertyes from dict
* support old dict like syntax in those classes
* remove describe_counts and describe_generic (logic moved to classes VarDescription)


vorel99:
missing spark refactoring (tbd)

fabclmnt:
Hi @vorel99 ,

we don't use the refactor. Please ensure that you use "feat" instead. 
In what concerns the proposed changes we will be analyzing and validating wether they fit the roadmap for the package.

PR #1449, opened 2023-09-18
fix: remove the duplicated cardinality threshold under categorical and text settings

ricardodcpereira:
The cardinality threshold used to distinguish between categorical and text features was duplicated in the `TextVars` and `CatVars` settings. The one from `TextVars` was removed, and the percentage cardinality threshold was moved to `CatVars`. Therefore, these configurations are now all under the `CatVars` settings.


alexbarros (approved):
Looks good to me. The failing tests does not seem to be related to this issue

PR #1460, opened 2023-09-26
Update formatters.py

ManinderAde12:


aquemy:
Hi @ManinderAde12 

Do you have an example on which the formatter fails? I would like to understand if there is a not a deeper issue here that the formatter.

fabclmnt:
@ManinderAde12 given the lack of response we will close this PR.

PR #1462, opened 2023-09-26
fix: table template ignoring text format

alexbarros:
closes https://github.com/ydataai/ydata-profiling/issues/1437
Update the table template to use the string in its original format without overflowing.

### DIFF
--- a/src/ydata_profiling/report/presentation/flavours/html/templates/table.html
+++ b/src/ydata_profiling/report/presentation/flavours/html/templates/table.html
@@ -18,11 +18,11 @@
                 <tr{% if 'alert' in row and row['alert'] %} class="alert"{% endif %}>
                     <th>{{ row['name'] }}</th>
                     {% if row['value'].__class__.__name__ == 'list' %}
-                        {% for value in row['value'] %}
-                            <td>{{ value }}{% if loop.last and 'hint' in row %} {{ row['hint'] }}{% endif %}</td>
+                        {% for value in row['value'] +%}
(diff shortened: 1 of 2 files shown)


aquemy (approved):

src/ydata_profiling/report/presentation/flavours/html/templates/table.html
```diff
@@ -18,11 +18,11 @@
                 <tr{% if 'alert' in row and row['alert'] %} class="alert"{% endif %}>
                     <th>{{ row['name'] }}</th>
                     {% if row['value'].__class__.__name__ == 'list' %}
-                        {% for value in row['value'] %}
-                            <td>{{ value }}{% if loop.last and 'hint' in row %} {{ row['hint'] }}{% endif %}</td>
+                        {% for value in row['value'] +%}
```
AmIIris, line 21: I think the '+' sign should not be there? I'm recieving the error 'TemplateSyntaxError: unexpected 'end of statement block'' when running this code.

gariciodaro: This breaking the render to html profile...

Ic3fr0g: It's true, removing it fixed the HTML render

alexbarros: This error was fixed in https://github.com/ydataai/ydata-profiling/pull/1491 and will be added to the next release

PR #1440, opened 2023-09-13
fix: bug when creating a new report

chrimaho:
Fix bug when creating a new report.

Caused by the `@typechecked` decorator on the `ProfileReport` class, and missing the `Optional[]` typehint on the `config_file` parameter.

Resolves #1439

PR #1470, opened 2023-10-03
chore(deps): update dependency typeguard to typeguard>=4.1.2, <5

maciejb:
4.1.5 as originally suggested by renovatebot fails because it drops Python 3.7 support. 4.1.2 is the last version with Python 3.7 support.

See my comment here https://github.com/ydataai/ydata-profiling/pull/1324#issuecomment-1736435984


maciejb:
@alexbarros It is my understanding that renovatebot is responsive to a change like this, however, I'm not sure about this. If you would rather I make an independent PR directly to `develop` please let me know.

maciejb:
The Multi-software tests appear to be failing because of the type check issue that was resolved in #1440.
@alexbarros Would you be able to rebase #1324 to the current `develop` branch?

Issue #1476, opened 2023-10-11
Column names should be like count, percent, sum as Y-data profiling might use these names. So if such names are used, ambiguity is created.


frelion:
Column names should be like count, percent, sum as Y-data profiling might use these names. So if such names are used, ambiguity is created.
My issue got resolved when I rename column name from 'count' to 'patient_count'.

_Originally posted by @waghts95 in https://github.com/ydataai/ydata-profiling/issues/1402#issuecomment-1663477461_
            
Can I solve this problem only by changing the column name?

fabclmnt:
Hi @frelion ,

not sure if I got the question.  Can you please provide more details?

frelion:
> Hi @frelion ,
> 
> not sure if I got the question. Can you please provide more details?

you can try running run this code below:
```
from ydata_profiling import ProfileReport
from pyspark.sql import SparkSession
from pyspark import SparkConf

SPARK_CONF = SparkConf()\
    .setMaster("local[2]")\
    .setAppName("test_ydata_profiling")
spark = SparkSession.builder.config(conf=SPARK_CONF).getOrCreate()

df = spark.createDataFrame([("A", 1), ("B", 2), ("C", 3)], ["catg", "count"])
ProfileReport(df)
```
error info:
```
---------------------------------------------------------------------------
AnalysisException                         Traceback (most recent call last)
File [~/anaconda3/envs/python3.8/lib/python3.8/site-packages/IPython/core/formatters.py:344](https://vscode-remote+wsl-002bubuntu-002d22-002e04.vscode-resource.vscode-cdn.net/home/sunzibin/tmp/~/anaconda3/envs/python3.8/lib/python3.8/site-packages/IPython/core/formatters.py:344), in BaseFormatter.__call__(self, obj)
    [342](https://vscode-remote+wsl-002bubuntu-002d22-002e04.vscode-resource.vscode-cdn.net/home/sunzibin/tmp/~/anaconda3/envs/python3.8/lib/python3.8/site-packages/IPython/core/formatters.py:342)     method = get_real_method(obj, self.print_method)
    [343](https://vscode-remote+wsl-002bubuntu-002d22-002e04.vscode-resource.vscode-cdn.net/home/sunzibin/tmp/~/anaconda3/envs/python3.8/lib/python3.8/site-packages/IPython/core/formatters.py:343)     if method is not None:
[...]

PR #1518, opened 2023-12-09
fix: Resolve the issue of conflicts between columns added during the analysis process and the original data columns in the Spark version.

frelion:
#1476 

In the Spark version, the program will add some auxiliary columns to the dataframe during runtime, such as Count, Std, etc. 

If the original data to be analyzed already contains these columns, it may result in column name conflicts.

solution:
Before the program analysis, add the suffix "_customer" to all columns of the DataFrame. 
Remove the suffix when displaying the results.


PeterlitsZo (commented):
Look good to me. It's OK. It can solve the problem.

fabclmnt (approved):

PR #1521, opened 2023-12-28
feat: support for python 3.12

aquemy:


.github/workflows/tests.yml
```diff
@@ -81,6 +85,7 @@ jobs:
           ${{ runner.os }}-${{ matrix.pandas }}-pip-
     - run: |
         pip install --upgrade pip setuptools wheel
+        virtualenv --upgrade-embed-wheels
```
Liam-Deacon, line 88: The CI is failing because virtualenv is not found, perhaps  try `pip install virtualenv && virtualenv --upgrade-embed-wheels` instead

portellaa (approved):

PR #1557, opened 2024-03-03
chore(deps): update python upper version

sagarishere:


alexbarros:
this is duplicated in https://github.com/ydataai/ydata-profiling/pull/1521

PR #1570, opened 2024-04-03
chore(deps): remove seaborn upper limit

ggydush:
Seaborn 0.13 fixes an issue with annotating heatmaps and I'd like to be able to use ydata-profiling in a repository that needs this fix.

I checked git blame and didn't see a reason for the upper-bound pin. I tested on 0.13.2 with Python 3.10 and all unit tests passed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ phik>=0.11.1,<0.13
 requests>=2.24.0, <3
 # Progress bar
 tqdm>=4.48.2, <5
-seaborn>=0.10.1, <0.13
+seaborn>=0.10.1
 multimethod>=1.4, <2
 # metrics
 statsmodels>=0.13.2, <1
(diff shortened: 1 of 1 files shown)


fabclmnt (changes requested):
Hi @ggydush ,

thank you for your contribution. Nevertheless, and due to past experiences where new minor releases from packages like numpy, seaborn or matplotlib have broken the functionality of ydata-profiling we have decided to analyze case by case the versions that we support. 

We can consider upgrading seaborn, but we will need to define an upper version.

ggydush:
Would you be okay with upper pin on latest (0.13.2) provided tests pass?

fabclmnt:
Yes totally! That would do.

fabclmnt (approved):

ggydush:
Thank you for approving! Is there any chance this could be included in the next release?

alexbarros:
@fabclmnt pode fazer o merge o erro do lint message era um erro do actions que jah foi corrigido

fabclmnt:
@ggydush yes this will be included for the next release. Probably either this week or beginning of the next.

Thank you for your contribution.

Issue #1466, opened 2023-09-27
multi language support


YIKAILucas:
### Missing functionality

The whole web page is English, and cannot display other language natively.

### Proposed feature

Could we have a multi-language support profiling? The community can support the ability to add more languages

whxue0807:
You can try this, add a .ttf file to replace the default path :        
def _plot_word_cloud(
    series: Union[pd.Series, List[pd.Series]],
    figsize: tuple = (6, 4),
) -> plt.Figure:
    if not isinstance(series, list):
        series = [series]
    plot = plt.figure(figsize=figsize)
    for i, series_data in enumerate(series):
        word_dict = series_data.to_dict()
        wordcloud = WordCloud(
            background_color="white", random_state=123, width=300, height=200, scale=2
**,font_path='/home/STFangSong.ttf'**
        ).generate_from_frequencies(word_dict)

        ax = plot.add_subplot(1, len(series), i + 1)
        ax.imshow(wordcloud)
        ax.axis("off")

    return plot

PR #1598, opened 2024-05-31
feat: Add font_path arg for wordcloud

DongHyun-JO:
This PR is a proposed solution to issue #1466.
The default font in Wordcloud is `DruidSansMono.ttf`, and if the languages ​​of the values ​​are different, the image may be broken.

* Example data (language: korean)
image
image

This is a known solution when using wordcloud, but the problem could not be solved because the 'font_path' argument was not specified in ydata_profiling.

I added a setting variable to 'Settings.plot' to pass the value to the font_path argument of word cloud.
Actually, I'm not sure if it's appropriate to put it here.
I just thought that it was a variable for wordcloud plot rendering, and the criterion for judgment was that I thought it could be used in other matplotlib plots as well.

If you do the following, you can check that the wordcloud is displayed normally.
```
profile_report = ProfileReport(df, title="test", minimal=True, plot={"font_path": "/Users/user/Downloads/Nanum_Gothic/NanumGothic-Regular.ttf"})
```

image

### DIFF
--- a/src/ydata_profiling/config.py
+++ b/src/ydata_profiling/config.py
@@ -173,6 +173,7 @@ class Plot(BaseModel):
     histogram: Histogram = Histogram()
     scatter_threshold: int = 1000
     cat_freq: CatFrequencyPlot = CatFrequencyPlot()
+    font_path: Optional[Union[Path, str]] = None
 
 
 class Theme(Enum):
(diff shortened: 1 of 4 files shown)


fabclmnt:
Hi @DongHyun-JO ,

thank you for your contribution! This is really helpful. 

The team will be reviewing the PR shortly.

alexbarros (approved):
:+1: thanks for the contribution

PR #1601, opened 2024-06-07
fix: timeseries_heatmap

tpuk0:
`timeseries_heatmap` from `ydata_profiling.visualisation.plot` drops an error, when 'entity_column' is of type str.
The reason for this is that the arguments to `pivot_table` are passed in the wrong order - see [documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.pivot_table.html). 

Existing unit test does not cover a case where entities are strings, thus the bug was missed.


fabclmnt:
HI @tpuk0 ,

thank you for your contribution! The team will be reviewing your PR shortly :)

alexbarros (approved):

PR #1619, opened 2024-07-12
fix: No module named 'pkg_resources'

### LINKED ISSUE #1618: ModuleNotFoundError: No module named 'pkg_resources'
### Current Behaviour

Thanks for the great repository!

Currently fails on the import with

```python
import ydata_profilling
```

Resulting in:

```
Traceback (most recent call last):
  File "c:\Users\foo\test.py", line 9, in <module>
    import ydata_profiling
  File "C:\Users\foo\AppData\Local\pypoetry\Cache\virtualenvs\foo-nnQxNQNi-py3.12\Lib\site-packages\ydata_profiling\__init__.py", line 14, in <module>
    from ydata_profiling.compare_reports import compare  # isort:skip # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\foo\AppData\Local\pypoetry\Cache\virtualenvs\foo-nnQxNQNi-py3.12\Lib\site-packages\ydata_profiling\compare_reports.py", line 12, in <module>
    from ydata_profiling.profile_report import ProfileReport
  File "C:\Users\foo\AppData\Local\pypoetry\Cache\virtualenvs\foo-nnQxNQNi-py3.12\Lib\site-packages\ydata_profiling\profile_report.py", line 9, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

### Expected Behaviour

The library should be successfully imported.

### Data Description

Currently not using any data as it fails on the import.

```Python
import ydata_profiling.
```

### pandas-profiling version

4.8.3

### Dependencies

```Text
annotated-types==0.7.0
attrs==23.2.0
certifi==2024.7.4
charset-normalizer==3.3.2
colorama==0.4.6
contourpy==1.2.1
cycler==0.12.1
dacite==1.8.1
fonttools==4.53.1
h5py==3.11.0
htmlmin==0.1.12
idna==3.7
ImageHash==4.3.1
Jinja2==3.1.4
joblib==1.4.2
kiwisolver==1.4.5
llvmlite==0.43.0
MarkupSafe==2.1.5
matplotlib==3.8.4
multimethod==1.12
networkx==3.3
numba==0.60.0
numpy==1.26.4
packaging==24.1
pandas==2.2.2
patsy==0.5.6
phik==0.12.4
pillow==10.4.0
pydantic==2.8.2
pydantic_core==2.20.1
pydicom==2.4.4
pyparsing==3.1.2
python-dateutil==2.9.0.post0
pytz==2024.1
PyWavelets==1.6.0
PyYAML==6.0.1
requests==2.32.3
scipy==1.13.1
seaborn==0.13.2
setuptools==70.3.0
six==1.16.0
statsmodels==0.14.2
tqdm==4.66.4
typeguard==4.3.0
typing_extensions==4.12.2
tzdata==2024.1
urllib3==2.2.2
visions==0.7.6
wordcloud==1.9.3
ydata-profiling==4.8.3
```

### OS

Windows 11

### Checklist

- [X] There is not yet another bug report for this issue in the [issue tracker](https://github.com/ydataai/pandas-profiling/issues)
- [X] The problem is reproducible from this bug report. [This guide](http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) can help to craft a minimal bug report.
- [X] The issue has not been resolved by the entries listed under [Common Issues](https://pandas-profiling.ydata.ai/docs/master/pages/support_contrib/common_issues.html).

abdrysdale:
Fixes #1618 add `setuptools` as a requirement.


fabclmnt (changes requested):
Hi @abdrysdale ,

thank you for your contribution. Nevertheless, it is odd that installing the `setuptools` is the solution for the issue. The behaviour that you are getting seems to be related with something more as we are not being able to replicate. 

Here is a screenshot of a conda environment that I've just created and where I've installed ydata-profiling. 

image

I would prefer to avoid adding the dependency to `setuptools` given this is something that is only needed to be installed in case you are building python yourself. 

I would put it as something optional in case the user is aware it is missing it in its environment.

fabclmnt:
Also, can you please make sure that your PR follows the naming and convention so that we are able to build the versions of the package automatically (eg. fix: (message), feat: (message), etc.)

abdrysdale:
I've managed to confirm this on my Windows PC by running the following:

```python

def test():

    # Ensure setuptools has not been installed alongside ydata_profiling
    try:
        import setuptools
        print("Setup tools was already installed - it shouldn't be!")
        assert False
        
    except ModuleNotFoundError as exception:
        assert str(exception) == "No module named 'setuptools'"

    # This fails with: ModuleNotFoundError: No module named 'pkg_resources'
    import ydata_profiling
        

if __name__ == "__main__":
    test()
```

Which fails with:

```
  File "C:\Users\foo\AppData\Local\pypoetry\Cache\virtualenvs\ydata-test-w9_z196s-py3.12\Lib\site-packages\ydata_profiling\profile_report.py", line 9, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

abdrysdale:
> Nevertheless, it is odd that installing the setuptools is the solution for the issue

How so? `pkg_resources` is part of `setuptools` (https://setuptools.pypa.io/en/latest/pkg_resources.html) so if `setuptools` isn't installed then `pkg_resources` won't be installed.

Can you confirm that `setuptools` isn't installed globally or elsewhere and that's why you're not getting the same error? Have you tried in a VM?

jklaise:
@fabclmnt I suggest reproducing using a Python 3.12 venv as from Python 3.12 `setuptools` is no longer bundled, see https://docs.python.org/3/whatsnew/3.12.html:
> [gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3/library/venv.html#venv-explanation) virtual environment.

Note that conda virtual envs might still bundle setuptools so the issue might not be apparent if using conda with Python 3.12. In short, imo adding `setuptools` as an explicit dependency for `python>=3.12` seems to be the right solution until the library migrates from `pkg_resources` to `importlib.resources`.

CesarArroyo09:
Please accept this and close/reject #1698.

As explained by @ssiegel in #1698, if something from `setuptools` is used within the library and we already know that there are some environments where `setuptools` is not present by default then `setuptools` should be added as an explicit dependence of the package.

CesarArroyo09:
I also noticed there is not `pyproject.toml` file in this project and it stills support Python versions that reached end-of-life. I would volunteer in upgrading the package management if I somebody would guide me through the process (as with feedback).

Anyway, great project and I hope it still has a lot of life ahead.

fabclmnt:
Hi @CesarArroyo09,

Thank you for taking the initiative! However, that action has [already been addressed by another community member](https://github.com/ydataai/ydata-profiling/pull/1448). While our team hasn't had the chance to review it yet due to time constraints, it will be merged in the next release.

Regarding _setuptools_, as mentioned previously, it is not required during ydata-profiling runtime, so I would not include it in the requirements for several reasons:
- Unnecessary in Many Environments – Many Python environments (e.g., venv, pipx) already install _setuptools_ automatically, making its inclusion redundant.
- Potential Version Conflicts – Specifying setuptools as a dependency could introduce compatibility issues, especially if a specific version is required.
- Additionally, since pyproject.toml has been in our backlog and is already in development, I don’t see the changes in this PR as necessary.

Thanks again for your contributions, and whenever you fell like contributing feel free to do so, the team will be happy to guide ans support you through the process.

fabclmnt:
Solved by #1711

fabclmnt:
Not needed. Issue reported solved by #1711

fabclmnt:
Not needed. Issue reported solved by https://github.com/ydataai/ydata-profiling/pull/1711

fabclmnt:
Not needed. Issue reported solved by #1711

abdrysdale:
> Regarding _setuptools_, as mentioned previously, it is not required during ydata-profiling runtime, so I would not include it in the requirements for several reasons:

It is required in runtime. I'm not, nor are other people with this issue, building Python from source. This fails when trying to import the module.
 
> * Unnecessary in Many Environments – Many Python environments (e.g., venv, pipx) already install _setuptools_ automatically, making its inclusion redundant.

Just because a few (even many) Python environments bundle setuptools doesn't mean the inclusion should be redundant. By nature of not including setuptools in the requirements they are no longer requirements.

> * Potential Version Conflicts – Specifying setuptools as a dependency could introduce compatibility issues, especially if a specific version is required.

Good dependency managers (such as poetry or uv) handle this for you. Testing will tell you what range of values are acceptable for your project and then if others have compatibility issues then these are indeed compatibility issues as both projects require different versions of setuptools. This package requires setuptools, you're just not specifying it in the requirements which doesn't solve the underlying issue.
 
> * Additionally, since pyproject.toml has been in our backlog and is already in development, I don’t see the changes in this PR as necessary.

setuptools can be added to the dependencies specified by the pyproject.toml file. Better yet would be to move to importlib.resources

gsouf:
works on python 3.11. Broken on 3.12+

PR #1632, opened 2024-07-30
fix: ignore None type categories

desobolevsky:
Actual PR for https://github.com/ydataai/ydata-profiling/issues/1325
Duplicate for https://github.com/ydataai/ydata-profiling/pull/1416, but since this PR seems to be outdated and no longer update/rebased to the latest version, doing this.

### DIFF
--- a/src/ydata_profiling/report/structure/variables/render_categorical.py
+++ b/src/ydata_profiling/report/structure/variables/render_categorical.py
@@ -136,6 +136,9 @@ def render_categorical_unicode(
     for category_alias_name, category_alias_counts in sorted(
         summary["category_alias_char_counts"].items(), key=lambda x: -len(x[1])
     ):
+        if category_alias_name is None:
+            continue
+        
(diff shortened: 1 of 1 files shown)


desobolevsky:
Hey, @fabclmnt! Don't wanna be intrusive, but could you please take a look at my PR? :)
The tests seem to fail for some external reasons not connected to changes at the code.
Thanks in advance!

alexbarros:
sry for the delay @desobolevsky, I will be having a look at this week. My questions with this PR is more related to the root cause of the problem, which is still not clear. This workaround may lead to silent errors, since the user is expecting their unicode categories to be categorized as `Other Symbol`, just like 😀 and 🔥.

Also, when creating a fix for specific use case like this we expect the addition of a unit test `tests/issues/test_issue1325.py` to ensure the issue is fixed and we don't run into it again in the future.

fabclmnt (approved):

PR #1649, opened 2024-09-17
fix: Correction for numpy.NaN deprecating to numpy.nan

joshicola:
Correction for the error:

`AttributeError: `np.NaN` was removed in the NumPy 2.0 release. Use `np.nan` instead.`


fabclmnt (approved):
🚀

PR #1651, opened 2024-09-25
fix: update describe_numeric_pandas.py

sregy:
* fix np.NaN -> np.nan


fabclmnt:
Hi @sregy ,

thank you for your contribution. Nevertheless this PR is a duplicate of another one that was previously reviewed and approved. 

This is great initiative but I do recommend that you check the history of issues , selected backlog for development and more importantly already open PRs so you don't have to invest your time with duplicate developments.

PR #1664, opened 2024-10-21
docs: fix broken link for doc for great_expectations.md

rugk:
I know it's deprecated (which is also a little bit too subtle to be documented IMHO), but the link should work, at least.


fabclmnt (changes requested):
Hi @rugk ,

thank you for your contribution. Nevertheless your commit messages do not comply with the expected format (eg feat: (description) or fix: (description)).

Please to contribute to our package make sure that you comply with that naming and convention. For now I've forced the change.

rugk:
Ah I thought the format you've followed was https://www.conventionalcommits.org/, which is why I instinctively used `docs:` (though it would also be hard to say whether this is a fix, yeah probably).

Unfortunately, that requirement is also not written on https://github.com/ydataai/ydata-profiling/blob/develop/CONTRIBUTING.md only can only see it when it's too late aka when a PR/commit is made… (in the CI or so)

Okay the CI still says:
>  Error: You have commit messages with errors

But not which one?
And I see you used `docs`  actually, so I was on the right track (for a different PR maybe), [and the CI requires this(?)](https://github.com/ydataai/ydata-profiling/blob/919553162d316670c2d9ca5e15cbfd7179ada73a/commitlint.config.js#L2), but it does not seem to work?

rugk:
Okay maybe it was the line length… anyway, it's a bit frustrating for me to guess what the CI actually wants from me…

fabclmnt:
> Okay maybe it was the line length… anyway, it's a bit frustrating for me to guess what the CI actually wants from me… (Edit: ah and now the CI does not run because of that GitHub workflow approval ha, oh well… I hope it is fine for you now.)

Based on what I see from the error from the linter, the error message indicates that your subject line starts with a capital letter, which is considered "sentence-case" (like a normal sentence) or "pascal-case" (where each word starts with a capital letter), which is not allowed by the linter's configuration.

Message from the linter error. 
```
input: docs: Fix broken link for doc for great_expectations.md

I know it's deprecated (which is also a little bit too
subtle to be documented IMHO), but the link should work, at least.
✖   subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
```

Ideally the history of commits needs to be updated in order to remove any upper case. 

Thank you for your contribution !! :)

PR #1676, opened 2024-11-16
feat: Add option to remove default stopwords from word summary 

### LINKED ISSUE #1655: Feature Request
### Missing functionality

Word clouds contain the most comon words, and for free text fields, these words are often: 'and', 'to', 'the', 'from' etc. Which provide no meaningful insight into the data.

### Proposed feature

Include an option to remove stopwords when generating word clouds. Perhaps by incorporating the `nltk` package to identify the list of stopwords.

Example:
```
from nltk.corpus import stopwords

stop = stopwords.words('english')
```

### Alternatives considered

Removing stopwords from a pandas dataframe prior to generating the report wouldn't suffice in this case, as it would affect the 'samples' in the report

### Additional context

_No response_

jabahm:
This PR introduces a new feature to provide greater flexibility in word summary analysis by adding the following enhancements:

- **Option to Remove Default Stopwords**: A new parameter `remove_default_stopwords` allows users to enable or disable the removal of default English stopwords.
- **Selective Stopword Retention**: Added `keep_stopwords` parameter to retain specific stopwords even when `remove_default_stopwords` is enabled.

Closes #1655

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,3 +9,4 @@ sphinx_rtd_theme>=0.4.3
 sphinx-autodoc-typehints>=1.10.3
 sphinx-multiversion>=0.2.3
 autodoc_pydantic
+nltk
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -8,3 +8,4 @@ ipython<9
 pyarrow
 twine>=3.1.1
 kaggle
+nltk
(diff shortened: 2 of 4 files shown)


fabclmnt (changes requested):
Hi @jabahm ,

thank you for you contribution!
This functionality can be achieve by using the stopwords config already available in the package. 
Perhaps you could share a bit more on what is the improvement that you expect this new development will bring, given the existing functionality. 

Regarding the implementation, that are certain additions that I would review or remove in order to eliminate complexity.

PR #1679, opened 2024-12-05
fix: type schema not checking for empty columns

alexbarros:
Type schema did not check for empty columns leading to errors when trying to calculate some metrics over it. Also updates the type of an unsupported column when the type was defined by the user.
e.g. for columns:
```python
df = pd.DataFrame({
    'date': ['0001-01-01', '9999-12-31', '2012-10-03', '2017-10-03', '2022-10-03', '2022-10-04'],
    'null': [None]*6,
})
```
image


fabclmnt (approved):
Looking good 🚀

PR #1704, opened 2025-02-21
add note about Python 3.13 support

afeld:


fabclmnt:
Hi @afeld ,

thank you for contributing for the package, nevertheless this is a change we won't accept as this would require more for the docs maintenance (information duplicated in the readme). 

The python versions supported by the package are clearly described in PyPi, Github (side bar) with no need for an explicit reference in the readme file. Furthermore, in case the package is not yet available for a certain python version the installation process error informs correctly the user and details thoroughly the versions that are supported.

PR #1674, opened 2024-11-07
Support Pandas future.infer_string=True in report generation

ssiegel:
Previously, report generation encountered issues when `future.infer_string=True` was set. This resulted in multiple warnings ("FutureWarning: Dtype inference on a pandas object is deprecated") and failures when string columns contained only empty strings ("AttributeError: 'StringDtype' object has no attribute 'pyarrow_dtype'").
This change resolves the issue by explicitly setting the dtype to "object" for the relevant operations.

PR #1707, opened 2025-02-28
fix: Fix CI pipeline

ssiegel:
I wouldn't normally expect this to be my responsibility to fix, but given the state of the project, I felt compelled to step in so that we can finally move forward with other PRs like #1619, #1672, #1673, and #1674. It's disappointing to see a company-backed project in this condition, with crucial issues left unaddressed for so long. At the very least, this should help clean things up and make it easier to take CI results seriously again.


ssiegel:
The current test failures are caused by IPython 9 (IPython being a transitive dependency of `nbval` in requirements-test.txt), something #1709 tries to fix. A quick workaround for the tests might be to add a `ipython<9` constraint into requirements-test.txt. Should I add that?

portellaa (approved):

PR #1716, opened 2025-03-05
fix: add information to ydataprofiling

fabclmnt:


sleeperservices:
Why is this being added to the __init__.py of your package. It spams my log 4 times when I call another package, fastparquet, that I use. I know you want to promote your other/new package but this is the wrong way to go about it.

There is a typo on line 29 so the h is missing from https. Why is it even here as you have the right url on line 8 and your not even using that line of the constructor.

what are these characters doing in this text"\033[1;34m{self.title}\033[0m" on line 53.

PR #1800, opened 2025-12-12
fix: Multiple Spark Enhancements

MCBoarder289:
This is a single PR that addresses multiple issues at once:

* https://github.com/ydataai/ydata-profiling/issues/1723
* https://github.com/ydataai/ydata-profiling/issues/1722
* https://github.com/ydataai/ydata-profiling/issues/1602
* https://github.com/ydataai/ydata-profiling/issues/1429

I'm closing my other PRs (see list below) because once I got to resolving each issue individually, other rendering issues cropped up, so it's easier to test all of the fixes at once. 

**Main Goal**
Combining all of the previous issues in Spark, it all boils down to the fact that the Spark reports were inaccurate when compared with Pandas. So my main goal here was to make Spark closer to parity with the Pandas output

**Example of pandas output on toy dataset:**
Pandas_example_stats

Pandas_example_common_values

**Example of *broken/pre-fixes* spark output on toy dataset:**
Spark_wrong_stats

Spark_wrong_common_values

**Fixed Spark Output**
Spark_fixed_stats_new

Spark_fixed_common_values_new

**Issues and Root Causes**
There are couple of commits in here that address specific root causes of these discrepancies. Here are the summarized issues with their solutions:

* **Issue 1:** pandas by default will count "NaN" values as Null in summary stats, but Spark SQL does not, so we explicitly address that in one of the commits.
    * **Resolution:** This was resolved by ensuring the numeric_stats_spark() method explicitly filters out nulls and Nans to match pandas' default behavior

* **Issue 2:** Missing values were not being properly calculated because NaN in Spark is not null, so they weren't considered missing when they should be
    * **Resolution:** Adding nan filters to the n_missing computation in the describe_spark_counts()method

* **Issue 3:** Histogram counts and Common Values counts using the summary["value_counts_without_nan"] Series were not correctly summing counts.
    * **Resolution:** Adding a sum to the counts, and removing the limit(200) makes everything line up to parity with the Pandas output

    * **NOTE:** Since we're pre-aggregating data for the value_counts, I don't think the limit(200) is necessary even with Spark. Since we're pulling this down into a Pandas Series anyway, if the data was too big, then that would explicitly fail the process instead of producing misleading reports. If you're running this in Spark, you're likely using a machine that has a good bit of memory anyway. 

**Concluding Thoughts**
While there is still some very slight variation to the computed stats because of how Spark handles nulls/NaNs differently than pandas, I think this new output is acceptably close to the pandas version and any differences are ultimately negligible. Especially when comparing the initial outputs where the differences are misleading without these fixes, or reports would not even complete/render with some edge cases (all null numeric columns, etc.)

@fabclmnt - Apologies for all of the tags, and I'm still open to all feedback on this approach! I'm happy to discuss further, and hope this is helpful to anyone using the Spark backend.

**Misc. Notes:**
* All pytest tests should be passing
* List of previous PRs that I'm closing in favor of this one:
    * https://github.com/ydataai/ydata-profiling/pull/1795
    * https://github.com/ydataai/ydata-profiling/pull/1796
    * https://github.com/ydataai/ydata-profiling/pull/1797
    * https://github.com/ydataai/ydata-profiling/pull/1798


MCBoarder289:
Hi @fabclmnt, thank you for all of your feedback here!

I've force-pushed all of my changes to this branch which I believe should address all of those issues. A brief summary of what I updated:

* Updated all of the commit messages to match the project’s commit convention.
* Many of the mypy issues stemmed from me making the histogram functions return None if they were invalid. I've wrapped those calls in a new utility function that will render an empty container in place of an image in those scenarios. Needed to update more report rendering files, but I feel like those should be safe and caught by unit tests.
* Moved all of the new issue tests that I added into the `tests/backends/spark_backend` directory.

Can you please take a look and let me know if there's anything else that needs to be updated?

Thanks again for your time!

fabclmnt:
Sorry about the time taken for the release, but here you can see ydata-profiling with the latest changes: http://pypi.org/project/ydata-profiling/#history

PR #1861, opened 2026-09-04
Fixes #637: Added a schema.json file

### LINKED ISSUE #637: Document fields in the JSON report

Combo1:
This PR adds a schema.json. 

The file is not used to verify the json file creation, but is only for documentation purposes, such that the user knows what a field means.

https://github.com/Data-Centric-AI-Community/fg-data-profiling/issues/637

### DIFF
--- a/src/data_profiling/schema/schema.json
+++ b/src/data_profiling/schema/schema.json
@@ -1,0 +1,131 @@
+{
+  "title": "YData Profiling Report Schema",
+  "description": "A documented schema describing the structure and meaning of fields in a ydata-profiling JSON report.",
+  "type": "object",
+  "properties": {
+    "analysis": {
+      "type": "object",
+      "description": "Title of the Profile Report and start date and end date of report generation",
+      "properties": {
+        "title": { "type": "string", "description" : "Title for the Profile Report"},
+        "date_start": { "type": "string", "description" : "Timestamp indicating when the profiling analysis started, in 'YYYY-MM-DD HH:MM:SS.ssssss' format"},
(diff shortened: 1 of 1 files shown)


fabclmnt (changes requested):
Hi @Combo1, thanks for contributing and choosing and open issue. I do have a couple of comments though. 

Fundamentally, the PR doesn't say why this shape of fix. #637 asks for
documentation of the JSON report fields, which means that content belongs in
docs/reference/ as a page on the JSON report, not a JSON file somewhere in the library.

On the schema itself:
- The structure doesn't match the report. `variables` is a top-level
  field of BaseDescription, a sibling of `table`, not nested inside it.
  `time_index_analysis` is a TimeIndexAnalysis object or None, not a
  string. See src/data_profiling/model/description.py.
- It covers 3 of the 11 top-level fields. scatter, correlations,
  missing, alerts, package, sample and duplicates are absent.
- Nothing validates it. As written it's a hand-maintained copy of the
  serialization that will drift silently. If we keep a schema, it needs
  a test asserting a real to_json() payload conforms to it.

If you want to take this on properly: a docs/reference page describing
the JSON fields, generated from or checked against the dataclasses in
model/description.py, would close #637.

